PR #8, opened 2022-09-26
Feat/write

BIRD311:


src/Lib/BSP/Lumps/BspLumps/EntityLump.cs
```diff
@@ -27,6 +27,7 @@ private bool ReadEntity(BinaryReader reader)
                 var inSection = false;
                 var inString = false;
                 char x;
+                //while (reader.BaseStream.Position < reader.BaseStream.Length
```
Gocnak, line 30: Leftover?

src/Lib/BSP/Lumps/BspLumps/EntityLump.cs
```diff
@@ -112,28 +113,42 @@ private bool ReadEntity(BinaryReader reader)
...
         {
             foreach (var ent in Data)
             {
-                writer.Write("{");
-                foreach (var (key, value) in ent.Properties)
-                    writer.Write($"\"{key}\" \"{value}\"");
-                foreach (var (key, value) in ent.IOProperties) 
-                    writer.Write($"\"{key}\" \"{value.TargetEntityName},{value.Input},{value.Parameter},{value.Delay},{value.TimesToFire}\"");
-                writer.Write("}\0");
+                stream.Write(Encoding.ASCII.GetBytes("{\n"));
```
Gocnak, line 138: How long does `Encoding.ASCII.GetBytes` take to execute? Should the static ones (all but the `${prop}` one) be cached outside of this foreach loop?

src/Lib/BSP/Lumps/GameLumps/StaticPropLump.cs
```diff
@@ -0,0 +1,241 @@
...
+            writer.Write(prop.FadeMinDist);
+            writer.Write(prop.FadeMaxDist);
+            writer.Write(prop.LightingOrigin.X);
+            writer.Write(prop.LightingOrigin.Y);
+            writer.Write(prop.LightingOrigin.Z);
+            // since v5
+            if (ActualVersion >= StaticPropVersion.V5)
+                writer.Write(prop.ForcedFadeScale);
+            // v6, v7, and v7* only
+            switch (ActualVersion)
```
Gocnak, line 188: No default case?

src/Lib/BSP/Struct/Entity.cs
```diff
@@ -57,11 +75,12 @@ public Entity(IEnumerable<KeyValuePair<string, string>> keyValues)
...
                 else
                 {
-                    Properties.Add(key, value);
+                    Properties.Add(new Property<string>(key, value));
                 }
             }
 
-            if (ClassName is null) Console.WriteLine("Warning: Found entity with missing classname! Fuck you!");
+            if (ClassName is null)
+                Console.WriteLine("Warning: Found entity with missing classname! Fuck you!");
```
Gocnak, line 83: :P

```suggestion
            if (ClassName is null)
                Console.WriteLine("Warning: Found entity with missing classname!");
```

Gocnak (changes requested):
Great stuff, really the only things I have issues with are the exception strings (and some other nitpicks). Could the exception strings be made better to better help devs / users figure out what's going wrong? Same goes for the console writing, which will be directly visible by the users of the commandline version.

src/Lib/BSP/Lumps/BspLumps/EntityLump.cs
BIRD311: I would hope the compiler/optimization understand that the output of this is always the same and not call this function every time. In fact it shouldn't be called at all during runtime since the ASCII value for a fixed string shouldn't change (but maybe you could override something and it can't do this at compile time for some reason).
Even if it calls it every time it shouldn't take long. I can check if you want to but I wouldn't worry about this for now. There are a lot more performance improvements to be made like asynchronous file reading and processing lumps in parallel. Not saying small things like this aren't important :)

src/Lib/BSP/Lumps/GameLumps/StaticPropLump.cs
BIRD311: yes it's just like all the other ifs there .. it's a switch so have to check 3 things

src/Lib/BSP/IO/LumpWriter.cs
```diff
@@ -0,0 +1,93 @@
...
+            else
+            {
+                mem.Seek(0, SeekOrigin.Begin);
+                compressedLength = mem.Length;
+                w.Write(LzmaId);
+                w.Write((int)uncompressedStream.Length);
+                w.Write((int)mem.Length - headerSize);
+                w.Write(lzmaStream.Properties);
+                if (w.BaseStream.Position != headerSize)
+                    throw new InvalidDataException("something something bad lzma header");
```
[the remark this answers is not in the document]
BIRD311: it should write exactly X amount of bytes here and checks against my hard coded number X so this exception shouldn't show up unless during development or maybe on some target system where ints aren't 32 bit (is that a thing in c#?)
More of an assert I guess.
Changed to you version of the message.

BIRD311:
The Console.WriteLine stuff I would want to replace with some sort of logging and I'm not sure what we would even keep for a cli version (with different logging levels we could keep most of it as debug output) but yes, I can go over my exceptions/messages and make them more useful (and not "math is hard" :) )

src/Lib/BSP/Lumps/BspLumps/EntityLump.cs
Gocnak: I completely agree, I'm fine with this for now!

src/Lib/BSP/Lumps/GameLumps/StaticPropLump.cs
Gocnak: Then it can be condensed to one if statement if it's just these three coalescing into one function call.

src/Lib/BSP/Struct/Entity.cs
tsa96: I take full responsibility for this

src/Lib/BSP/IO/BspFileWriter.cs
```diff
@@ -0,0 +1,104 @@
...
+                if ((lump is GameLump || lump is PakFileLump) && lump.Compress)
+                {
+                    Console.WriteLine($"let's not compress {lump.GetType().Name} .. it's a silly place");
+                    lump.Compress = false;
+                }
+                LumpHeader newHeader = Write(lump);
+                LumpHeaders[lumpType] = new BspLumpHeader(newHeader, lump.Version);
+                Console.WriteLine($"Lump {lumpType}({(int)lumpType})\n\t{newHeader.Offset}\n\t{newHeader.Length}");
+            }
+        }
```
tsa96, line 72: There should definitely be a newline between these two methods. Again, this is a nit, I apologise, but worth keeping clean

src/Lib/BSP/IO/BspFileWriter.cs
```diff
@@ -0,0 +1,104 @@
...
+                var lump = _bsp.Lumps[lumpType];
+                if (!lump.Empty())
+                {
+                    //Lump offsets (and their corresponding data lumps) are always rounded up to the nearest 4-byte boundary, though the lump length may not be. 
+                    int padTo = (lumpType == BspLumpType.Physlevel) ? 16 : 4;
+                    var pad = new byte[(padTo - BaseStream.Position % padTo)];
+                    if (pad.Length != padTo)
+                        Write(pad);
+                    startPosition = (int)BaseStream.Position;
+                }
```
tsa96, line 62: For readability I really think we should break stuff up with newlines a bit more. Certainly between one if statement closing and another openings, having the newline helps show they're not logically connected as an `if/else` is.

src/Lib/BSP/IO/BspFileWriter.cs
```diff
@@ -0,0 +1,104 @@
...
+                LumpHeaders[lumpType] = new BspLumpHeader(newHeader, lump.Version);
+                Console.WriteLine($"Lump {lumpType}({(int)lumpType})\n\t{newHeader.Offset}\n\t{newHeader.Length}");
+            }
+        }
+        private void ConstructTexDataLumps()
+        {
+            // TODO: check obeys source limits
+            List<string> texStrings = new();
+
+            var texdata = _bsp.GetLump<TexDataLump>().Data;
```
tsa96, line 78: camelCase

src/Lib/BSP/IO/BspFileReader.cs
```diff
@@ -0,0 +1,210 @@
...
+            if (_bsp.Lumps.Any())
+                throw new InvalidDataException("bsp lumps and reader headers didn't match!");
+            _bsp.Lumps = newLumps;
+
+        }
+
+        //for testing
+        private bool CheckOverlapping()
+        {
+            bool ret = false;
```
tsa96, line 138: It's a nit I know, but I'd prefer we be more consistent with `var` vs explicit types. For something that's so clearly a `bool` (from the RHS) I think we should `var` in these cases.

src/Lib/BSP/IO/BspLumpHeader.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public BspLumpHeader(LumpHeader header, int version)
+        {
+            Offset = (int)header.Offset;
+            Length = (int)header.Length;
+            Version = version;
+            FourCc = header.Compressed
+                        ? (int)header.UncompressedLength
+                        : 0;
+        }
```
tsa96, line 13: newline

src/Lib/BSP/IO/GameLumpHeader.cs
```diff
@@ -0,0 +1,31 @@
...
+        { }
+
+        public GameLumpHeader(LumpHeader tmpHeader, ushort version, int id)
+        {
+            Id = id;
+            Fileofs = (int)tmpHeader.Offset;
+            //set last bit to 0 (only seen flags as 1 and 0 so far but its probably a bitfield)
+            Flags &= 0xFFFE;
+            //length is always the uncompressed length
+            Filelen = (int)tmpHeader.UncompressedLength;
```
tsa96, line 15: var names

tsa96: I really think these matter here since they're properties. So `FileLen` and `FileOfs` (only capitalise 2-letter initials)

src/Lib/BSP/IO/GameLumpReader.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public void Load()
+        {
+            Lumps.Clear();
+            _gameLump.Lumps.Clear();
+            ReadHeader();
+            LoadAll();
+        }
+
+        protected override void ReadHeader()
```
tsa96, line 31: I know I'm being so annoying about this, but there's not a single newline throughout this entire function, I really think it harms the readability.

src/Lib/BSP/IO/LumpHeader.cs
```diff
@@ -0,0 +1,29 @@
...
+            Offset = offset;
+            UncompressedLength = uncompressedLength;
+            CompressedLength = compressedLength;
+
+        }
+        public long Offset { get; set; }
+        public long UncompressedLength { get; set; }
+        public long CompressedLength { get; set; }
+
+        public bool Compressed { get { return CompressedLength >= 0; } }
```
tsa96, line 25: Can be `public bool Compressed => CompressedLength >= 0;`

src/Lib/BSP/IO/LumpHeader.cs
```diff
@@ -0,0 +1,29 @@
...
+        //the actual length
+        public long Length { get { return Compressed ? CompressedLength : UncompressedLength; } }
```
tsa96, line 27: As above

src/Lib/BSP/IO/LumpReader.cs
```diff
@@ -0,0 +1,101 @@
...
+                if (decompress)
+                {
+                    MemoryStream decompressedStream = Decompress();
+                    lumpReader = new BinaryReader(decompressedStream);
+                    lumpStreamLength = decompressedStream.Length;
+                    if (lump is IUnmanagedLump unmanagedLump)
+                    {
+                        unmanagedLump.Compressed = false;
+                        unmanagedLump.UncompressedLength = -1;
+                    }
```
tsa96, line 73: If `decompress` is true then `lump` is not `IUnmanagedLump`. So this is always false

src/Lib/BSP/IO/LumpReader.cs
```diff
@@ -0,0 +1,101 @@
...
+                }
+                else if (lump is IUnmanagedLump unmanagedLump)
+                {
+                    unmanagedLump.Compressed = true;
+                    unmanagedLump.UncompressedLength = lumpHeader.UncompressedLength;
+                    lumpReader = this;
+                    lumpStreamLength = lumpHeader.CompressedLength;
+                }
+                else
```
tsa96, line 82: Similarly, because `decompress` iff `lump is not IUnmanagedLump`, `if (decompress)` and `else if (lump is IUnmanagedLump unmanagedLump)` is exhaustive - this can never be reached.

src/Lib/BSP/IO/LumpReader.cs
```diff
@@ -0,0 +1,101 @@
...
+                    throw new NotImplementedException(
+                        "not decompressing a managed lump not implemented .."
+                        + " make it unmanaged if you don't care about it or decompress and read");
+            }
+            else
+            {
+                if (lump is IUnmanagedLump unmanagedLump)
```
tsa96, line 89: Looks like we should reuse the `decompress` bool here. Though I think it would be clear to name that variable something like `shouldDecompress`.

src/Lib/BSP/Lumps/Lump.cs
```diff
@@ -1,20 +1,32 @@
...
+    public abstract class Lump<T> : Lump
+                          where T : System.Enum
+    {
+        public T Type { get; set; }
+        protected Lump(BspFile parent) : base(parent)
+        { }
+    }
+
+    // implements how a lump will be read from a stream (same for write)
+    // doesn't store offset or length because header information is serparate
```
tsa96, line 15: separate

src/Lib/BSP/Lumps/GameLumps/StaticPropLump.cs
```diff
@@ -0,0 +1,241 @@
...
+            if (ActualVersion >= StaticPropVersion.V11)
+                prop.UniformScale = System.BitConverter.ToSingle(reader.ReadBytes(4));
+            Data.Add(prop);
+            if (reader.BaseStream.Position - startpos != StructureSize)
+                throw new InvalidDataException($"StaticProp structuresize doesn't match reader position after read ({reader.BaseStream.Position - startpos} != {StructureSize})");
+        }
+
+        protected override void WriteItem(BinaryWriter writer, int index)
+        {
+            long startpos = writer.BaseStream.Position;
```
tsa96, line 160: https://upload.wikimedia.org/wikipedia/commons/thumb/4/43/07._Camel_Profile%2C_near_Silverton%2C_NSW%2C_07.07.2007.jpg/1200px-07._Camel_Profile%2C_near_Silverton%2C_NSW%2C_07.07.2007.jpg

src/Lib/BSP/Lumps/UnmanagedLump.cs
```diff
@@ -1,12 +1,56 @@
...
+
+        public override void Read(BinaryReader reader, long length)
         {
+            DataStream = reader.BaseStream;
+            DataStreamOffset = reader.BaseStream.Position;
+            DataStreamLength = length;
         }
+        public override void Write(Stream stream)
+        {
+            if (DataStream != null)
```
tsa96, line 33: Can early return here

tsa96 (changes requested):
Overall this is great, your class structures make far more sense than what I was doing. Most of these comments are nits, but I think for a new project like this it's worth paying attention to code cleanliness. Generally I just think more newlines breaking up the program logic into clearer chunks helps readability a ton.

src/Lib/BSP/IO/BspFileReader.cs
```diff
@@ -0,0 +1,210 @@
...
+            }
+
+            _bsp.Revision = ReadInt32();
+
+            UpdateGameLumpLength();
+
+            SortLumps();
+
+            if (CheckOverlapping())
+                throw new InvalidDataException("overlapping lumps bad");
```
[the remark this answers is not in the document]
BIRD311: lost my exception message changing in this file during rebase apparently :/

src/Lib/BSP/IO/BspFileReader.cs
```diff
@@ -0,0 +1,210 @@
...
+                    }
+                    else if (header.Offset > prevEnd)
+                    {
+                        Console.WriteLine($"space between lumps {prevLump.Type} {prevEnd} <-- {header.Offset - prevEnd} --> {header.Offset} {lump.Type}");
+                    }
+                    if (header.Offset + header.Length >= prevEnd)
+                    {
+                        prevLump = lump;
+                        prevHeader = header;
+                    }
```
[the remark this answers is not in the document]
BIRD311: I disagree. The condition affects those 2 lines. A continue affects the entire loop. Sure it would work here but if anyone comes back to this code an wants to add something later they would need think about why the loop should stop here and than probably change it to a if again. Also the only reason you don't have to indent is because you hide the continue in the same line as the if. I don't think this change makes it easier to read and I don't like one line if statements unless its as simple as "if(skip) continue;"

src/Lib/BSP/IO/BspLumpHeader.cs
BIRD311: Above this you mean? Because it needs a check and its 3 friends don't?

src/Lib/BSP/IO/GameLumpReader.cs
BIRD311: Wrong, there are 2 and half of them should be removed :)
Yes I will add a few.

src/Lib/BSP/IO/LumpReader.cs
BIRD311: Yes but only because "decompress" is hard coded to check the type right now. Not sure why we would need to decompress a unmanaged lump in the future and not make it a managed type instead but in theory you can decompress them lumps as well so that's why its here. I would keep it.

src/Lib/BSP/IO/LumpReader.cs
BIRD311: I need the unmanagedLump variable here so I thought I just make it like this instead of casting and if we would change how decompress gets its value this would break.
Why is decompress not clear?. I don't mind changing it but then I should also change "Compress" in the lump class to "ShouldCompress" and "Compressed" to "IsCompressed" so its consistent, right?

src/Lib/BSP/IO/BspFileReader.cs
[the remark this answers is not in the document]
tsa96: Okay, fair enough, I suppose doesn't actually read better the way I was suggesting

src/Lib/BSP/IO/LumpReader.cs
tsa96: Mmh okay. My issue is just that doing this `is IUnmanagedLump unmanagedLump` multiple times, storing the bool in one case but then doing the same check again without using the stored bool, it makes the overall logic a little confusing. I personally like using language that makes it clear where things are bools but that's really just preference I guess.

In this case I think just removing the `decompress` bool is best, it's only used once. But maybe I'm being overly nitpicky about it, once the unreachable code is gone it should read much better anyway

src/Lib/BSP/IO/BspLumpHeader.cs
tsa96: Oh, I meant between the constructor and properties, helps to keep them distinct

src/Lib/BSP/IO/LumpReader.cs
tsa96: If it could be useful in the future then you could comment it out but leave it, but I don't think it's right to have code that's logically impossible to be true in uncommented code. At the very least leave a comment to say what you just said, we don't want other programmers reading the code left scratching their heads over blocks of code that make no sense logically

BIRD311:
everything should be done now

Gocnak (approved):

tsa96 (approved):

PR #10, opened 2022-12-01
Misc cleanup, polish

sapphonie:
- Adds Pastel for console colorization
- Adds more error spew if something goes wrong
- Remove dup comments/dead code
- Add barebones usage section
- Changed syntax, lumper will now accept an infile bsp and dump an outfile bsp specified on console


BIRD311:
Since the old Program.cs is gone now, should be just close this or do you want to add some of it to the Lumper.UI? The args stuff would be nice in the UI project.

sapphonie:
I'll go ahead and stick it in UI

PR #17, opened 2023-02-07
Remove Escape Characters from File Path

patrickmoc:
Lumper was not loading files properly due to an invalid file path being used when initializing the bsp file object. For instance loading arena_offblast_final would result in the file path:
`C:\Program%20Files%20(x86)\Steam\steamapps\common\Team%20Fortress%202\tf\maps\arena_offblast_final.bsp`
The file would fail to open (on windows) with the escape characters in place of the spaces. The fix for this is to strip these to provide 
`C:\Program Files (x86)\Steam\steamapps\common\Team Fortress 2\tf\maps\arena_offblast_final.bsp`
using C#'s Uri features.


tsa96 (approved):
Yup, fix works for me (also on Windows).

PR #19, opened 2023-02-28
Implement Drag-n'-Drop support

### LINKED ISSUE #18: Allow opening BSP files by dragging over UI view
Only looked at this briefly but I think it's possible (on Windows at least) (https://github.com/AvaloniaUI/Avalonia/issues/2635). We want users, especially moderators reviewing maps, to be able to open BSP files very quickly to check contents - if someone's reviewing a dozen maps in one sitting we should make it very streamlined.

Best way I can think of is having dragging a BSP file anywhere onto the UI immediately open that BSP. If another BSP file (by file path) is already open, spawn a new instance of Lumper and open in there. Should be lightweight enough to handle that, though maybe worth checking someone doesn't accidentally drag 100 files onto it and spawn a 100 instances at once.

Riven-Spell:
Resolves #18

This is probably on the hacky/bad side, due to my general lack of familiarity with Avalonia-- But, I'm happy to hear any and all feedback, even if it's "Start from scratch and do it this way". I'm here to learn and contribute.

### DIFF
--- a/src/Lumper.UI/App.axaml.cs
+++ b/src/Lumper.UI/App.axaml.cs
@@ -1,5 +1,6 @@
 using Avalonia;
 using Avalonia.Controls.ApplicationLifetimes;
+using Avalonia.Input;
 using Avalonia.Markup.Xaml;
 using Lumper.UI.ViewModels;
 using Lumper.UI.Views;
@@ -17,10 +18,12 @@ public class App : Application
     {
         if (ApplicationLifetime is IClassicDesktopStyleApplicationLifetime
             desktop)
+        {
             desktop.MainWindow = new MainWindow
             {
                 DataContext = new MainWindowViewModel()
             };
+        }
 
         base.OnFrameworkInitializationCompleted();
     }
--- a/src/Lumper.UI/ViewModels/MainWindowViewModel.IO.cs
+++ b/src/Lumper.UI/ViewModels/MainWindowViewModel.IO.cs
@@ -163,7 +163,7 @@ public partial class MainWindowViewModel
         if(!file.TryGetUri(out var path))
         {
            throw new Exception("Failed to get file path");
-            
+
         }
         LoadBsp(path.AbsolutePath);
     }
--- a/src/Lumper.UI/ViewModels/MainWindowViewModel.cs
+++ b/src/Lumper.UI/ViewModels/MainWindowViewModel.cs
@@ -1,6 +1,10 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using Avalonia;
 using Avalonia.Controls.ApplicationLifetimes;
+using Avalonia.Input;
 using Lumper.UI.ViewModels.Bsp;
 using ReactiveUI;
 
@@ -27,6 +31,11 @@ public partial class MainWindowViewModel : ViewModelBase
         SearchInit();
         TabsInit();
         IOInit();
+
+        if (desktop.Args?.Length > 0)
+        {
+            LoadBsp(desktop.Args[0]);
+        }
     }
 
     public IClassicDesktopStyleApplicationLifetime Desktop
@@ -44,5 +53,49 @@ public partial class MainWindowViewModel : ViewModelBase
     {
         get => _selectedNode;
         set => this.RaiseAndSetIfChanged(ref _selectedNode, value);
+    }
+
+    internal static void DragOver(object? sender, DragEventArgs e)
+    {
+        e.DragEffects = DragDropEffects.Link;
+
+        var names = e.Data.GetFileNames() ?? new List<string>();
+
+        if (!e.Data.Contains(DataFormats.FileNames) || !names.FirstOrDefault("").ToLower().EndsWith(".bsp"))
+            e.DragEffects = DragDropEffects.None;
+    }
+
+    internal void Drop(object? sender, DragEventArgs e)
+    {
+        if (!e.Data.Contains(DataFormats.FileNames))
+            return;
+
+        var names = e.Data.GetFileNames() ?? new List<string>();
+        foreach (string target in names)
+        {
+            if (!target.ToLower().EndsWith(".bsp"))
+                continue;
+
+            if (BspModel == null)
+            {
+                // if nothing is open, open it.
+                LoadBsp(target);
+            }
+            else
+            {
+                // Otherwise, open a brand new Lumper instance with it.
+                string? executableFileName = Process.GetCurrentProcess().MainModule?.FileName;
+                if (executableFileName == null)
+                    return;
+
+                ProcessStartInfo startInfo = new()
+                {
+                    ArgumentList = { $"{target}" },
+                    FileName = executableFileName,
+                };
+
+                Process.Start(startInfo);
+            }
+        }
     }
 }
--- a/src/Lumper.UI/Views/MainWindow.axaml.cs
+++ b/src/Lumper.UI/Views/MainWindow.axaml.cs
@@ -1,12 +1,28 @@
 using System;
 using System.ComponentModel;
 using Avalonia.Controls;
+using Avalonia.Input;
 using Lumper.UI.ViewModels;
 
 namespace Lumper.UI.Views;
 
 public partial class MainWindow : Window
 {
+    // Quietly shadow DataContext to add initializer
+    public new object? DataContext
+    {
+        init // DataContext is never overwritten-- It's ok to not defensively remove the Drop handler, and do DragOver here.
+        {
+            var dc = value as MainWindowViewModel; // It must be (as of 3/16/23)
+            AddHandler(DragDrop.DropEvent, dc!.Drop);
+            AddHandler(DragDrop.DragOverEvent, MainWindowViewModel.DragOver);
+
+            base.DataContext = value;
+        }
+
+        get => base.DataContext;
+    }
+
     public MainWindow()
     {
         InitializeComponent();
--- a/src/Lumper.UI/Views/MainWindow.axaml
+++ b/src/Lumper.UI/Views/MainWindow.axaml
@@ -14,6 +14,7 @@
         TransparencyLevelHint="AcrylicBlur"
         ExtendClientAreaToDecorationsHint="True"
         WindowStartupLocation="CenterScreen"
+        DragDrop.AllowDrop="True"
         Background="Transparent">
 
     <Design.DataContext>


BIRD311:
Did unfortunately not work for me on linux (tested i3 and openbox). Maybe you could test this again on you system. The linked issue mentions opening new lumper instanced for multiple files.

Riven-Spell:
> Did unfortunately not work for me on linux (tested i3 and openbox). Maybe you could test this again on you system.

Works fine for me on Windows-- I've yet to try on my Linux install, but I'll get around to it after work today.

> The linked issue mentions opening new lumper instanced for multiple files.

I can look into doing this.

Riven-Spell:
> I'll get around to it after work today.

Liar, liar, pants on fire!

> Did unfortunately not work for me on linux (tested i3 and openbox). Maybe you could test this again on you system.

https://github.com/AvaloniaUI/Avalonia/issues/6085 Appears to be a known issue. That's fine for now.

> The linked issue mentions opening new lumper instanced for multiple files.

Trying to get this working, being a bit finnicky.

tsa96 (changes requested):
Implementation looks okay, works for me on Windows. We definitely need to show some indicator when stuff in loading but that's a separate thing and requires async refactor that BIRD is writing up an issue on now.

Then a few general housekeeping issues:
* Your editor seems to have changed a bunch of line ending or something, meaning loads of lines of changes that shouldn't be there.
* Your branch includes merge commits, which we don't allow. This is our fault really, we don't have good guidelines for this repo. I've been drafting some general contribution guidelines this afternoon that'll be on the organisation README in the future. In the meantime, here's a preview:
  - Keep your branch up-to-date by rebasing, not using merge commits. We consider merge commits to be a major detriment to the quality of a Git repo, and will always ask you to remove any merge commits in code review, even if this means force-pushing.
  - Keep your commits clean and simple, splitting all changes into separate commits.
  - Don’t push commits specifically for addressing issues in earlier commits, including those brought up in code review. Instead, make those commits then fixup them into the original commit using an interactive rebase. Again, we have no problem with force-pushes; so long as you make sensible local backups, we feel that rewriting history of non-master branches is a necessary evil to avoid merge commits. For a general guide on this workflow, see [here](https://medium.com/singlestone/a-git-workflow-using-rebase-1b1210de83e5).
  - Name your commits using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) (no Gitmojis!). If in doubt, just check the rest of the history of the repo and copy that.

It's probably easiest to address these changes, then just `reset --mixed` and recommit everything.

src/Lumper.UI/Views/MainWindow.axaml.cs
```diff
@@ -1,21 +1,74 @@
...
+    {
+        if (e.Data.Contains(DataFormats.FileNames))
+        {
+            var names = e.Data.GetFileNames() ?? new List<String>();
+            foreach (string target in names)
+            {
+                if (!target.ToLower().EndsWith(".bsp"))
+                    continue;
+
+                if (((MainWindowViewModel)DataContext!).BspModel == null)
```
BIRD311, line 50: There is too much logic and not enough UI here for it to be in `View`. Adding the drag and drop events in code-behind is fine if its easier but most of this can probably be moved to the `ViewModel`. At the very least the process spawning should be in the `ViewModel` in my opinion.
The ViewModel itself shouldn't know any avalonia specific things. (It does right now, but thats a me problem)

BIRD311 (changes requested):

src/Lumper.UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1,6 +1,10 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using Avalonia;
 using Avalonia.Controls.ApplicationLifetimes;
+using Avalonia.Input;
```
BIRD311, line 7: Do we need a `using Avalonia` here? I wanted to get them out of the ViewModel at some point. I think the ViewModel souldn't depend on the View.

src/Lumper.UI/ViewModels/MainWindowViewModel.IO.cs
```diff
@@ -148,13 +148,13 @@ private async void Save(string path)
         _bspModel.FilePath = path;
     }
 
-    private void LoadBsp(string path)
+    public void LoadBsp(string path)
```
BIRD311, line 151: I think this doesn't need to be public now.

BIRD311 (changes requested):

src/Lumper.UI/ViewModels/MainWindowViewModel.IO.cs
Riven-Spell: It does-- Still referenced in App.axaml.cs for the command-line argument.

What we *could* do is pass the arguments along into MainWindowViewModel and have that handle the opening of files/new instances.

src/Lumper.UI/ViewModels/MainWindowViewModel.cs
Riven-Spell: Unfortunately, DragEventArgs relies upon `Avalonia.Input`.

tsa96 (approved):

tsa96:
Looks good to me! @BIRD311 did you bring up the issue about the file location or whatever you were talking about yesterday?

src/Lumper.UI/App.axaml.cs
```diff
@@ -17,11 +18,22 @@ public override void OnFrameworkInitializationCompleted()
...
+        {
+            var vm = new MainWindowViewModel();
             desktop.MainWindow = new MainWindow
             {
-                DataContext = new MainWindowViewModel()
+                DataContext = vm,
             };
 
+            desktop.MainWindow.AddHandler(DragDrop.DragOverEvent, MainWindowViewModel.DragOver);
+            desktop.MainWindow.AddHandler(DragDrop.DropEvent, vm.Drop);
```
BIRD311, line 29: This can be in MainWindow.axaml.cs because it only adds events to the MainWindow.

src/Lumper.UI/App.axaml.cs
```diff
@@ -17,11 +18,22 @@ public override void OnFrameworkInitializationCompleted()
...
+
+            if (desktop.Args?.Length > 0)
+            {
+                vm.LoadBsp(desktop.Args[0]);
+            }
```
BIRD311, line 34: This can be in the MainWindowViewModel so you don't need to create a new ViewModel. There is a `desktop` already in the constructor in `MainWindowViewModel.cs` but maybe it should be in `MainWindowViewModel.IO.cs`. But we might add different `desktop.Args` in the future though so your choice.

src/Lumper.UI/ViewModels/MainWindowViewModel.IO.cs
BIRD311: looked at the `App` stuff again and I think it should just be in MainWindow*

src/Lumper.UI/ViewModels/MainWindowViewModel.cs
BIRD311: yeah didn't find a solution I liked. I _think_ the ViewModel should be reusable even if you change your View framework or whatever. Adding a function in the code-behind which takes the strings and passes them to the ViewModel doesn't sound like the way to go. For unit testing this is probably fine.

BIRD311 (changes requested):

src/Lumper.UI/App.axaml.cs
Riven-Spell: I'll pull this into VM for now. There's room for more robust argument parsing in a different PR.

src/Lumper.UI/App.axaml.cs
Riven-Spell: Correct-- Tl;dr the stuff in the brackets happens AFTER the constructor code. DataContext is not set until afterwards. Hence, I "hid" the DataContext parameter and used a initer/getter to configure things at init.

There is no setter present, but it does get the job done.

Riven-Spell: The brackets are roughly equivalent to

```csharp
var bar = new foo();
bar.x = 5;
bar.y = 6;
```

not

```csharp
bar.x = 5;
bar.y = 6;
var bar = new foo();
```

BIRD311: Still creates new objects in the App.axaml.cs though.

Riven-Spell: That's a one-line revert.

src/Lumper.UI/App.axaml.cs
```diff
@@ -1,5 +1,6 @@
 using Avalonia;
 using Avalonia.Controls.ApplicationLifetimes;
+using Avalonia.Input;
```
BIRD311, line 3: not needed

src/Lumper.UI/App.axaml.cs
```diff
@@ -17,10 +18,12 @@ public override void OnFrameworkInitializationCompleted()
     {
         if (ApplicationLifetime is IClassicDesktopStyleApplicationLifetime
             desktop)
+        {
```
BIRD311, line 21: just revert this file so the drag and drop commit only touches file it has to

src/Lumper.UI/ViewModels/MainWindowViewModel.IO.cs
```diff
@@ -163,7 +163,7 @@ private async Task LoadBsp(IStorageFile file)
         if(!file.TryGetUri(out var path))
         {
            throw new Exception("Failed to get file path");
-
```
BIRD311, line 166: same here, revert or move formatting stuff to a separate commit

BIRD311 (changes requested):

tsa96:
This could be picked up again once my PR is in, either by the original author if they're still around, or someone else. But I'd suggest just rewriting / copying stuff over rather than rebasing.

The DataContext shadowing stuff I don't understand, just use ReactiveWindow<MainWindowViewModel> (might be doing this alrwsy on my PR, don't remember) and WhenActivated. And setting up the event handlers should be done in the view.

Also, would be nice to alter the appearance of window slightly when dragging, or at least change the cursor.

tsa96:
Reworked this a bit on https://github.com/momentum-mod/lumper/pull/106, added you as co-author on the commit. Appreciate the contribution!

PR #22, opened 2023-04-08
Tasks UI

BIRD311:
basic UI so we can actually use the lumpertasks

### DIFF
--- a/src/Lumper.UI/Converters/TaskStatusConverter.cs
+++ b/src/Lumper.UI/Converters/TaskStatusConverter.cs
@@ -1,0 +1,51 @@
+using System;
+using System.Globalization;
+using Avalonia.Data;
+using Avalonia.Data.Converters;
+using Lumper.UI.ViewModels.Tasks;
+
+namespace Lumper.UI.Converters;
+
+/// <summary>
+///     format TaskStatus
+/// </summary>
+public class TaskStatusConverter : IValueConverter
+{
+    public static TaskStatusConverter Instance
+    {
+        get;
+    } = new();
+
+    public object Convert(object? value, Type targetType, object? parameter,
+        CultureInfo culture)
+    {
+        if (value is not TaskStatus status)
+        {
+            return new BindingNotification(
+                new ArgumentOutOfRangeException(nameof(value)),
+                BindingErrorType.DataValidationError);
+        }
+        switch (status)
+        {
+            case TaskStatus.Waiting:
+                return "|";
+            case TaskStatus.Running:
+                return "➠";
+            case TaskStatus.Success:
+                return "✓";
+            case TaskStatus.Failed:
+                return "⚠";
+            case TaskStatus.Unknown:
+            default:
+                return " ";
+        }
+    }
+
+    public object ConvertBack(object? value, Type targetType, object? parameter,
+        CultureInfo culture)
+    {
+        return new BindingNotification(
+            new ArgumentOutOfRangeException(nameof(value)),
+            BindingErrorType.DataValidationError);
+    }
+}
--- a/src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.Item.cs
+++ b/src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.Item.cs
@@ -1,0 +1,68 @@
+﻿using ReactiveUI;
+using Lumper.Lib.Tasks;
+
+namespace Lumper.UI.ViewModels.Tasks;
+
+/// <summary>
+///     ViewModel for ChangeTextrue Task
+/// </summary>
+public partial class ChangeTextureTaskViewModel : TaskViewModel
+{
+    public abstract class ChangeTextureItem : ViewModelBase
+    {
+        public ChangeTextureItem(ChangeTextureTaskViewModel parent,
+                                 ChangeTextureTask task)
+        {
+            Parent = parent;
+            _task = task;
+        }
+
+        ~ChangeTextureItem()
+        {
+            Remove();
+        }
+
+        public ChangeTextureTaskViewModel Parent { get; }
+
+        protected readonly ChangeTextureTask _task;
+
+        protected string _source = "";
+        public string Source
+        {
+            get => _source;
+            set
+            {
+                if (value != _source && value is not null)
+                {
+                    Remove();
+                    _source = value;
+                    Add(Source, Target);
+
+                    this.RaisePropertyChanged();
+                }
+            }
+        }
+
+        protected string _target = "";
+        public string Target
+        {
+            get => _target;
+            set
+            {
+                if (value != _target && value is not null)
+                {
+                    UpdateValue(value);
+                    _target = value;
+                    this.RaisePropertyChanged();
+                }
+            }
+        }
+
+        public abstract bool Add(string? source = null,
+                                 string? target = null);
+
+        protected abstract void Remove();
+
+        protected abstract void UpdateValue(string? newValue);
+    }
+}
--- a/src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.RegexItem.cs
+++ b/src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.RegexItem.cs
@@ -1,0 +1,52 @@
+﻿using System.Text.RegularExpressions;
+using System.Collections.Generic;
+using Lumper.Lib.Tasks;
+
+namespace Lumper.UI.ViewModels.Tasks;
+
+/// <summary>
+///     ViewModel for ChangeTextrue Task
+/// </summary>
+public partial class ChangeTextureTaskViewModel : TaskViewModel
+{
+    public class ChangeTextureRegexItem : ChangeTextureItem
+    {
+        public ChangeTextureRegexItem(ChangeTextureTaskViewModel parent,
+                                      ChangeTextureTask task,
+                                      KeyValuePair<Regex, string>? item = null)
+            : base(parent, task)
+        {
+            if (item is not null)
+            {
+                _source = item.Value.Key.ToString();
+                _target = item.Value.Value;
+            }
+        }
+
+        public override bool Add(string? source = null,
+                                 string? target = null)
+        {
+            source ??= Source;
+            target ??= Target;
+
+            _task.ReplaceRegex.Add(
+                new KeyValuePair<Regex, string>(new Regex(source), target));
+            return true;
+        }
+
+        protected override void Remove()
+        {
+            var idx = _task.ReplaceRegex.FindIndex(
+                x => x.Key.ToString() == Source
+                    && x.Value == Target);
+            if (idx > 0)
+                _task.ReplaceRegex.RemoveAt(idx);
+        }
+
+        protected override void UpdateValue(string? newValue)
+        {
+            Remove();
+            Add(Source, newValue);
+        }
+    }
+}
--- a/src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.cs
+++ b/src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.cs
@@ -1,0 +1,58 @@
+﻿using System;
+using System.Collections.ObjectModel;
+using Lumper.Lib.Tasks;
+
+namespace Lumper.UI.ViewModels.Tasks;
+
+/// <summary>
+///     ViewModel for ChangeTextrue Task
+/// </summary>
+public partial class ChangeTextureTaskViewModel : TaskViewModel
+{
+    public ChangeTextureTaskViewModel(ChangeTextureTask task)
+        : base(task)
+    {
+        try
+        {
+            foreach (var item in task.Replace)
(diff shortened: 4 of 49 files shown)


src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.Item.cs
```diff
@@ -0,0 +1,67 @@
...
+            get => _source;
+            set
+            {
+                if (value != _source && value is not null)
+                {
+                    Remove();
+                    _source = value;
+                    Add(Source, Target);
+
+                    this.RaisePropertyChanged(nameof(Source));
```
tsa96, line 40: From https://www.reactiveui.net/api/reactiveui/ireactiveobjectextensions/63fddd3b: "Leave null to let the runtime set to caller member name.". So you don't need the nameof, just this.RaisePropertyChanged() is fine.

src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.Item.cs
```diff
@@ -0,0 +1,67 @@
...
+        public string Target
+        {
+            get => _target;
+            set
+            {
+                if (value != _target && value is not null)
+                {
+                    UpdateValue(value);
+                    _target = value;
+                    this.RaisePropertyChanged(nameof(Target));
```
tsa96, line 55: Same as above

src/Lumper.UI/ViewModels/Tasks/TasksViewModel.cs
```diff
@@ -0,0 +1,175 @@
...
+        BspFile = bsp;
+        TaskTypes = new()
+        {
+            new TaskMenuItem(this, typeof(ChangeTextureTask)),
+            new TaskMenuItem(this, typeof(StripperTask)),
+            new TaskMenuItem(this, typeof(CompressionTask)),
+            new TaskMenuItem(this, typeof(RunExternalToolTask)),
+        };
+        this.WhenAnyPropertyChanged(nameof(SelectedTask))
+         .Subscribe(x => OnSelectedTaskChanged());
```
tsa96, line 33: Best to use `_` for unused params

src/Lumper.UI/ViewModels/Tasks/TasksViewModel.cs
```diff
@@ -0,0 +1,175 @@
...
+        if (SelectedTask is null || Tasks.Count <= 1)
+            return;
+        int idx = Tasks.IndexOf(SelectedTask);
+        int newIdx = idx + 1;
+        if (newIdx >= Tasks.Count)
+            newIdx = 0;
+        Tasks.Move(idx, newIdx);
+        SelectedTask = null;
+        SelectedTask = Tasks[newIdx];
+    }
```
tsa96, line 111: Can we just generalise these two to `MoveSelectedTask(int dir)`

src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.Item.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+        }
+
+        protected string _target = "";
+        public string Target
+        {
+            get => _target;
+            set
+            {
+                if (value != _target && value is not null)
```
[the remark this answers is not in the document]
tsa96: also early return

src/Lumper.UI/Converters/TaskStatusConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+                case TaskStatus.Failed:
+                    return "⚠";
+                case TaskStatus.Unknown:
+                default:
+                    return " ";
+            }
+        }
+        return new BindingNotification(
+            new ArgumentOutOfRangeException(nameof(value)),
+            BindingErrorType.DataValidationError);
```
tsa96, line 41: Just return the notif if `value is not TaskStatus status`

src/Lumper.UI/ViewModels/Tasks/TaskViewModel.cs
```diff
@@ -0,0 +1,81 @@
...
+    public bool _isRunning;
+    public bool IsRunning
+    {
+        get => _isRunning;
+        set => this.RaiseAndSetIfChanged(ref _isRunning, value);
+    }
+
+    public bool Run(BspFile bsp)
+    {
+        bool ret = true;
```
tsa96, line 56: Give clearer name like `success`, `wasSuccessful`, w/e

src/Lumper.UI/Views/Tasks/TasksView.axaml.cs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    private void InitializeComponent()
+    {
+        AvaloniaXamlLoader.Load(this);
+    }
+
+    private static IReadOnlyList<FilePickerFileType> GenerateJsonFileFilter()
+    {
+        var bspFilter = new FilePickerFileType("Json files");
```
tsa96, line 40: "JSON" not "Json"

src/Lumper.UI/Views/Tasks/TasksView.axaml.cs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public async void OnLoadClick(object sender, RoutedEventArgs e)
+    {
+        if (Desktop.MainWindow is null)
+            return;
+
+        var dialog = new FilePickerOpenOptions();
+        dialog.AllowMultiple = false;
+        dialog.Title = "Pick BSP file";
```
tsa96, line 55: Shouldn't this be "Load task file"?

src/Lumper.UI/Views/Tasks/TasksView.axaml.cs
```diff
@@ -0,0 +1,76 @@
...
+            return;
+        if (DataContext is TasksViewModel vm)
+            vm.Load(await result[0].OpenReadAsync());
+    }
+    public async void OnSaveClick(object sender, RoutedEventArgs e)
+    {
+        if (Desktop.MainWindow is null)
+            return;
+        var dialog = new FilePickerSaveOptions();
+        dialog.Title = "Pick BSP file";
```
tsa96, line 68: Ditto above

src/Lumper.UI/Views/Tasks/TasksView.axaml
```diff
@@ -0,0 +1,59 @@
...
+             mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
+             x:Class="Lumper.UI.Views.Tasks.TasksView">
+    <UserControl.Resources>
+        <converters:TaskStatusConverter x:Key="taskStatus" />
+        <converters:PercentConverter x:Key="percent" />
+    </UserControl.Resources>
+    <Grid RowDefinitions="Auto, *, Auto" ColumnDefinitions="*,2*">
+        <DockPanel Grid.Row="0" Grid.Column="0">
+            <Button DockPanel.Dock="Left" Click="OnLoadClick" IsEnabled="{Binding !IsRunning}">Load</Button>
+            <Button DockPanel.Dock="Left" Click="OnSaveClick" IsEnabled="{Binding !IsRunning}">Save</Button>
```
tsa96, line 17: Let's make it clearer what this is doing. We have horizontal space free, let's do "Load Tasks"/"Save Tasks"

tsa96 (changes requested):
Some comments on UI stuff. Styling doesn't matter for now but we should still have the layout in a decent shape.

- Up, Down ,Delete button should be on same row as Load, Save stuff
- The RMB to bring up Add... CM is really unintuitive, even for dev-facing tool this feels bad. I'd do "Add" button next to Up, Down, Delete
- Sidebar should be resizable like the node list in the browser view
Untitled_-_Tom_-_Obsidian_v1 1 1609,04,23,152343

BIRD311:
> Some comments on UI stuff. Styling doesn't matter for now but we should still have the layout in a decent shape.
> 
>     * Up, Down ,Delete button should be on same row as Load, Save stuff
> 
>     * The RMB to bring up Add... CM is really unintuitive, even for dev-facing tool this feels bad. I'd do "Add" button next to Up, Down, Delete
> 
>     * Sidebar should be resizable like the node list in the browser view
>       Untitled_-Tom-_Obsidian_v1 1 1609,04,23,152343

image
On your resolution maybe. This is usable but if I move the buttons to the top row it gets cut off. In the future I would like to get rid if the Up/Down and move it to a context menu on the menu item. Drag and drop to change the order would also be nice. The Delete can also go to the context menu. The Add context menu should probably not be a context menu at all. For now this was just an easy way to do and I don't want to spend to much time on this.

I can move the Up/Down/Delete row below the Load/Save so its closer or try to get it to wrap around to a new line if the window gets to small, but like I said, there should only be 3 buttons (Load, Save and Add)

src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.Item.cs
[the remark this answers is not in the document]
BIRD311: strings can be null

src/Lumper.UI/ViewModels/Tasks/TasksViewModel.cs
BIRD311: I don't get it .. whats unused?

src/Lumper.UI/Converters/TaskStatusConverter.cs
BIRD311: if I return early I have to cast again .. I like this more .. or what do you mean? `is *not* something *value*` doesn't make sense, does it?

src/Lumper.UI/ViewModels/Tasks/TasksViewModel.cs
tsa96: `x`

src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    public void AddRegexItem()
+    {
+        if (Task is ChangeTextureTask c)
+        {
+            var tmp = new ChangeTextureRegexItem(this, c);
+            if (tmp.Add())
+                RegexItems.Add(tmp);
```
tsa96, line 47: Oh sorry, shouldn've brought up the above thing with renaming `tmp` here as well

src/Lumper.UI/Converters/TaskStatusConverter.cs
tsa96: Try it, compiler makes the inference fine! i.e.
```c#
if (value is not TaskStatus status)
    return new BindingNotification(
        new ArgumentOutOfRangeException(nameof(value)),
        BindingErrorType.DataValidationError);
switch (status)
{
    case TaskStatus.Waiting:
        return "|";
    case TaskStatus.Running:
        return "➠";
    case TaskStatus.Success:
        return "✓";
    case TaskStatus.Failed:
        return "⚠";
    case TaskStatus.Unknown:
    default:
        return " ";
}
```

src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.Item.cs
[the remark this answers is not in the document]
tsa96: We have nullable reference types on, and `ChangeTextureReplaceItem` ctor takes KV-pair of non-nullable strings. If value can be null, should be KeyValuePair<string, string?>`

tsa96 (changes requested):

src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.RegexItem.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+
+        public override bool Add(string? source = null,
+                                 string? target = null)
+        {
+            if (source == null)
+                source = Source;
+            if (target == null)
+                target = Target;
```
tsa96, line 32: Crap, sorry, missed the same thing with `??=` here.

tsa96 (changes requested):

src/Lumper.UI/ViewModels/Tasks/ChangeTextureTaskViewModel.Item.cs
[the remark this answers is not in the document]
BIRD311: I'm just saying that `string s = null;` works, so `Target = null;` or `Source = null;` works .. I don't want `null` there so check for it .. the only reason I made some of the strings explicitly nullable is to get rid of the warnings you get when doing `string s = null;`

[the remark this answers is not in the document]
tsa96: Shouldn't you make the value in the kv pair string? then? In which case doesn't the compiler give you different NRT warnings?

tsa96 (approved):

PR #23, opened 2023-04-26
Feat/compare

BIRD311:


src/Lumper/BSP/JsonContractResolver.cs
```diff
@@ -0,0 +1,46 @@
...
+                    || x.PropertyName != nameof(IUnmanagedLump.DataStreamOffset));
+                ordered = ordered.Where(x =>
+                    (
+                       !x.DeclaringType.IsAssignableFrom(typeof(LumpHeader))
+                    && !x.DeclaringType.IsAssignableFrom(typeof(BspLumpHeader))
+                    && !x.DeclaringType.IsAssignableFrom(typeof(GameLumpHeader))
+                    )
+                    || x.PropertyName != nameof(LumpHeader.Offset));
+            }
+            return ordered.ToList();
```
tsa96, line 43: Really don't like this nesting, early returning here is better even if it means writing `ToList()` twice.

src/Lumper/BSP/Lumps/BspLumps/PakFileLump.cs
```diff
@@ -1,12 +1,34 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System.Security.Cryptography;
 
 namespace Lumper.Lib.BSP.Lumps.BspLumps
 {
+    //[JsonConverter(typeof(PakFileJsonConverter))]
     public class PakFileLump : ManagedLump<BspLumpType>
     {
+        [JsonIgnore()]
```
tsa96, line 16: Unnecessary parens here

src/Lumper/BSP/Lumps/Lump.cs
```diff
@@ -13,9 +15,13 @@ protected Lump(BspFile parent) : base(parent)
 
     // implements how a lump will be read from a stream (same for write)
     // doesn't store offset or length because header information is separate
+
+    [JsonConverter(typeof(JsonSubtypes), "Class")]
     public abstract class Lump
     {
+        public string Class { get => this.GetType().Name; }
         public bool Compress { get; set; }
+        [JsonIgnore()]
```
tsa96, line 24: Parens

src/Lumper/BSP/BspFile.cs
```diff
@@ -14,12 +15,14 @@ public class BspFile
         public const int HeaderSize = 1036;
         public const int MaxLumps = 128;
 
+
+        [JsonIgnore()]
```
tsa96, line 19: Unnecessary parens

src/Lumper.UI/Views/MainWindow.axaml
```diff
@@ -38,6 +38,7 @@
                         <Separator />
                         <MenuItem Header="_Save" Command="{Binding SaveCommand}" />
                         <MenuItem Header="_Save As..." Command="{Binding SaveAsCommand}" />
+                        <MenuItem Header="_ToJson..." Command="{Binding BspToJsonCommand}" />
```
tsa96, line 41: "Export JSON data"?

tsa96 (changes requested):

src/Lumper/BSP/JsonContractResolver.cs
BIRD311: this BIRD likes nesting .. and I don't like multiple returns .. and I don't like redundant code. I usually just change this stuff because I don't like arguing about it, but I really don't see the benefit in changing this.

tsa96: Okay fair enough, looking back at this you're totally right, it'd make the logic a lot less clear.

Though, I do find this block genuinely really hard to read. The indentation is pretty idiosyncratic but whatever, eventually we'll add some formatting rules that will keep everything consistent. I think the real problem is that the logic is just totally over-complicated. I had to copy this over to a text editor to pick it apart, and can't it just simplify down to this?

```c#
if (IgnoreOffset)
{
  ordered = ordered.Where(x =>
    !(x.DeclaringType.IsAssignableTo(typeof(IUnmanagedLump))
      || x.DeclaringType.IsAssignableFrom(typeof(LumpHeader))
      || x.DeclaringType.IsAssignableFrom(typeof(BspLumpHeader))
      || x.DeclaringType.IsAssignableFrom(typeof(GameLumpHeader))
      || x.PropertyName == nameof(IUnmanagedLump.DataStreamOffset)
      || x.PropertyName == nameof(LumpHeader.Offset)
    )
}
return ordered.ToList();
```

tsa96 (changes requested):

src/Lumper/BSP/JsonContractResolver.cs
BIRD311: the weird formatting is supposed to make it easier to read .. you code doesn't do the same thing .. how about I invert the entire thing instead?
```C#
                ordered = ordered.Where(x =>
                    !(x.DeclaringType.IsAssignableTo(typeof(IUnmanagedLump))
                    && x.PropertyName == nameof(IUnmanagedLump.DataStreamOffset)));
                ordered = ordered.Where(x =>
                    !((
                       x.DeclaringType.IsAssignableFrom(typeof(LumpHeader))
                    || x.DeclaringType.IsAssignableFrom(typeof(BspLumpHeader))
                    || x.DeclaringType.IsAssignableFrom(typeof(GameLumpHeader))
                    )
                    && x.PropertyName == nameof(LumpHeader.Offset)));
```

tsa96: Ohhhhh I'm sorry, I flipped the top one around wrong. But definitely using the DeMorgan rule or whatever like that makes it easy to follow. In terms of formatting, I just don't find the specific indentation you've done helpful, since you have to first deduce the reasoning behind the formatting method before you can tell what it's denoting. Again, with consistent formatting that'd be fine, but reviewing this I keep having to figure stuff out case-by-case.

tsa96: Sorry, to clarify, let's use the change you suggested

tsa96 (approved):

PR #28, opened 2023-07-08
Feat/pakfile UI

BIRD311:
good enough .. lets do the rest in review or leave some todos in


src/Lumper/BSP/Lumps/BspLumps/PakFileLump.cs
```diff
@@ -2,120 +2,180 @@
...
+                    continue;
+                if (existing.Count() == 1)
+                {
+                    Zip.RemoveEntry(existing.First());
+                }
+                else if (existing.Count() > 1)
+                {
+                    throw new NotImplementedException(
+                        "You can have multiple entries" +
+                        "with the same name in a zip?");
```
tsa96, line 84: Usually pakfiles are file-system based which afaik won't let you make duplicates, plus Source probably won't be happy if you have any. So even if the ZIP spec may formally support it (incl. https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.zipfileextensions.createentryfromfile?view=net-7.0 remarks), I don't think we should ever support it ourselves. So a `NotSupportedException` would be more appropriate, with an error like `$"Paklump has multiple entries of the same key: {existing.Key}".`

src/Lumper/BSP/Lumps/BspLumps/PakFileLump.cs
```diff
@@ -2,120 +2,180 @@
...
-        public override bool CanRead
+        public PakFileLump(BspFile parent) : base(parent)
         {
-            get { return false; }
+            Compress = false;
         }
     }
 
+    /*
+        public class ZipJsonConverter : JsonConverter<ZipArchive>
```
tsa96, line 141: Huh? If this is being left out temporarily best to leave a comment, otherwise delete. Really don't like having big blocks of random code sitting around.

src/Lumper.UI/ViewModels/Bsp/Lumps/PakFile/PakFileEntryVtfViewModel.cs
```diff
@@ -0,0 +1,327 @@
...
+        {
+            string err = VTFAPI.GetLastError();
+            Console.WriteLine(err);
+        }
+        _entry.DataStream.Seek(0, SeekOrigin.Begin);
+    }
+
+    private static Image<Rgba32> GetImageFromRgba8888(byte[] img, int width, int height)
+    {
+        var asdf = new Rgba32[width * height];
```
tsa96, line 285: "asdf"?

tsa96 (changes requested):
We'll definitely need to add to the UI in the future such as a way to quickly view every image in the pakfile, the UI here is mooostly okay, let's get it merged and do another pass in the future

src/Lumper.UI/ViewModels/Bsp/BspViewModel.Search.cs
```diff
@@ -41,5 +41,12 @@ private void SearchInit()
         //TODO: Add lock when search is slower than throttle rate
         var matcher = matcherBase.ConstructMatcher(pattern.Trim());
         await bspNode.Filter(matcher);
+        /*if (!string.IsNullOrEmpty(pattern))
```
Panzerhandschuh, line 44: Remove commented out code

src/Lumper.UI/ViewModels/Bsp/Lumps/PakFile/PakFileEntryBranchViewModel.cs
```diff
@@ -0,0 +1,203 @@
...
+    public PakFileEntryBranchViewModel CreatePathRoot(string entryKey, out string name)
+    {
+        return _pakFileViewModel.EntryRoot.CreatePath(entryKey, out name);
+    }
+
+    private PakFileEntryBranchViewModel CreatePath(string entryKey, out string name, int index = 0)
+    {
+        string[] path = entryKey.Split('/');
+
+        bool isDir;
```
Panzerhandschuh, line 111: This can be simplified:
`bool isDir = (index < path.Length - 1);`

Panzerhandschuh (approved):

tsa96 (approved):

PR #30, opened 2023-08-28
fix: set FilePath when creating BspViewModel

### LINKED ISSUE #20: Show BSP file name in window title
Currently there's no real indication what file you have open. We should probably show this in the UI itself somewhere but let's get through more UI design work first.

However, we should *definitely* modify the window title to like `bhop_arcane.bsp - Lumper`. I assume this is very easy, a good first issue!

wertiop121:
Closes #20.
This also fixes the save button opening the file picker instead of saving the currently opened file.


BIRD311:
thanks for the PR .. I'll get this in after the #29 PR because we can't save to the same file right now

tsa96:
Tweaked title naming a bit to improve window name and handle null cases

BIRD311 (approved):

Issue #3, opened 2022-08-30
Mounted Asset Removal


Gocnak:
We actually have a C++ tool shipping with the game that does this: "bspfix.exe"

So ideally all we need to do is invoke it on the bsp here.

tsa96:
For clarification, we have bspfix in C++ but want to write a C# version. Rewriting the C++ exactly would require license, but the concept is simple, compare a hash of every file in the Paklump with Valve assets of corresponding file type.

We don't want to require everyone running these checks have all content downloaded, instead we'll ship a JSON file with releases.

We need:
- Functionality mentioned above of traversing the Paklump and hash checking every file against a JSON list of hashes
- To be able to generate hashes of every file in a VPK and write to JSON list

Both should be usable via CLI and UI. It's important we can hook the hash generation up to our internal CI, as well as perform the hash checks via command-line, _maaaaaybe_ integrating into the backend eventually.

tsa96:
See https://discord.com/channels/235111289435717633/998769771451519088/1072388945146552370 for discussion of this

BIRD311:
Unassigned my self for now in case someone wants to pick this up. Calling bspfix on the map was implemented* but rewriting this in C# was not.

tsa96:
Oh -- closed by https://github.com/momentum-mod/lumper/pull/87!

PR #32, opened 2024-01-27
Add texture browser

ddorab:
resim

One issue with this is that it is a bit laggy if you scroll too fast and there's a lot of images. Avalonia doesn't have any virtualization for WrapPanel but we could implement [this](https://github.com/AvaloniaUI/Avalonia/pull/11251). They haven't merged it because of it's insufficient support for elements of different sizes but that's not an issue for us.

Also I've added the truncation because it's very ugly if we show really long names. You can see the full name when you hover over.

### DIFF
--- a/src/Lumper.UI/Models/Vtf/VtfFileData.cs
+++ b/src/Lumper.UI/Models/Vtf/VtfFileData.cs
@@ -1,0 +1,190 @@
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.PixelFormats;
+using VTFLib;
+using Lumper.Lib.BSP.Struct;
+
+namespace Lumper.UI.Models;
+
+public class VtfFileData
+{
+    private readonly PakFileEntry _entry;
+    private readonly uint _imageIndex;
+
+    private uint _frameCount;
+    private uint _faceCount;
+    private uint _mipmapCount;
+    private uint _depth;
+    private VTFImageFlag _flags;
+    private uint _majorVersion;
+    private uint _minorVersion;
+    private uint _imageSize;
+    private uint _imageWidth;
+    private uint _imageHeight;
+    private VTFImageFormat _imageFormat;
+
+    public uint FrameCount => _frameCount;
+    public uint FaceCount => _faceCount;
+    public uint MipmapCount => _mipmapCount;
+    public uint Depth => _depth;
+    public VTFImageFlag Flags => _flags;
+    public uint MajorVersion => _majorVersion;
+    public uint MinorVersion => _minorVersion;
+    public uint ImageSize => _imageSize;
+    public uint ImageWidth => _imageWidth;
+    public uint ImageHeight => _imageHeight;
+    public VTFImageFormat ImageFormat => _imageFormat;
+
+    static VtfFileData()
+    {
+        VTFAPI.Initialize();
+    }
+    public VtfFileData(PakFileEntry entry)
+    {
+        _entry = entry;
+        using var mem = new MemoryStream();
+        _entry.DataStream.CopyTo(mem);
+        byte[] vtfBuffer = mem.ToArray();
+
+        VTFFile.CreateImage(ref _imageIndex);
+
+        VTFFile.BindImage(_imageIndex);
+        VTFFile.ImageLoadLump(vtfBuffer, (uint)vtfBuffer.Length, false);
+
+        Update();
+    }
+
+    public Image<Rgba32>? GetImage(uint frame, uint face, uint slice, uint mipmapLevel)
+    {
+        VTFFile.BindImage(_imageIndex);
+
+        if (VTFFile.ImageGetHasImage() == 0)
+        {
+            return null;
+        }
+
+        IntPtr imageData = VTFFile.ImageGetData(frame, face, slice, mipmapLevel);
+        int imageSize = (int)VTFFile.ImageComputeImageSize(_imageWidth, _imageHeight, 1, 1, _imageFormat);
+
+        byte[] data = new byte[imageSize];
+        Marshal.Copy(imageData, data, 0, imageSize);
+
+        if (imageSize <= 0)
+            throw new ArgumentException("image data array size is 0");
+        byte[] dest = new byte[_imageWidth * _imageHeight * 4];
+        VTFFile.ImageConvertToRGBA8888(data, dest, _imageWidth, _imageHeight, _imageFormat);
+
+        var rgba = new Rgba32[_imageWidth * _imageHeight];
+        int j = 0;
+        for(int i = 0; i < dest.Length; i += 4)
+        {
+            rgba[j++] = new Rgba32(
+                dest[i],
+                dest[i + 1],
+                dest[i + 2],
+                dest[i + 3]);
+        }
+
+        return Image.LoadPixelData<Rgba32>(rgba.AsSpan(), (int)_imageWidth, (int)_imageHeight);
+    }
+
+    public void SetImageData(Image<Rgba32> image, uint frame, uint face, uint slice, uint mipmapLevel)
+    {
+        VTFFile.BindImage(_imageIndex);
+        byte[] imageRgbaData = GetRgba8888FromImage(image, out _);
+
+        int size = (int)VTFFile.ImageComputeImageSize(
+            (uint)image.Width, (uint)image.Height, 1, 1, _imageFormat);
+        byte[] vtfImageData = new byte[size];
+
+        VTFFile.ImageConvertFromRGBA8888(
+            imageRgbaData,
+            vtfImageData,
+            (uint)image.Width,
+            (uint)image.Height,
+            _imageFormat
+        );
+
+        VTFFile.ImageSetData(frame, face, slice, mipmapLevel, vtfImageData);
+        SaveVtf();
+    }
+
+    public void SetNewImage(Image<Rgba32> image)
+    {
+        VTFFile.BindImage(_imageIndex);
+        byte[] buffer = GetRgba8888FromImage(image, out bool hasAlpha);
+        var createOptions = new SVTFCreateOptions();
+        VTFFile.ImageCreateDefaultCreateStructure(ref createOptions);
+        createOptions.imageFormat =
+            hasAlpha ? VTFImageFormat.IMAGE_FORMAT_DXT5 : VTFImageFormat.IMAGE_FORMAT_DXT1;
+        if (!VTFFile.ImageCreateSingle(
+                (uint)image.Width,
+                (uint)image.Height,
+                buffer,
+                ref createOptions))
+        {
+            string err = VTFAPI.GetLastError();
+            Console.WriteLine(err);
+        }
+
+        SaveVtf();
+    }
+
+    private void SaveVtf()
+    {
+        uint size = VTFFile.ImageGetSize();
+
+        var vtfBuffer = new byte[size];
+        _entry.DataStream = new MemoryStream(vtfBuffer);
+
+        uint uiSize = 0;
+        if (!VTFFile.ImageSaveLump(vtfBuffer, (uint)vtfBuffer.Length, ref uiSize))
+        {
+            string err = VTFAPI.GetLastError();
+            Console.WriteLine(err);
+        }
+
+        _entry.DataStream.Seek(0, SeekOrigin.Begin);
+        Update();
+    }
+
+    private void Update()
+    {
+        _depth = VTFFile.ImageGetDepth();
+        _frameCount = VTFFile.ImageGetFrameCount();
+        _faceCount = VTFFile.ImageGetFaceCount();
+        _mipmapCount = VTFFile.ImageGetMipmapCount();
+        _flags = (VTFImageFlag)VTFFile.ImageGetFlags();
+        _majorVersion = VTFFile.ImageGetMajorVersion();
+        _minorVersion = VTFFile.ImageGetMinorVersion();
+        _imageSize = VTFFile.ImageGetSize();
+        _imageWidth = VTFFile.ImageGetWidth();
+        _imageHeight = VTFFile.ImageGetHeight();
+        _imageFormat = VTFFile.ImageGetFormat();
+    }
+
+    private static byte[] GetRgba8888FromImage(Image<Rgba32> image, out bool hasAlpha)
+    {
+        int size = image.Width * image.Height * 4;
+        using var mem = new MemoryStream();
+        var buffer = new byte[size];
+        int i = 0;
+        hasAlpha = false;
+        for (int y = 0; y < image.Height; y++)
+        {
+            for (int x = 0; x < image.Width; x++)
+            {
+                Rgba32 pixel = image[x, y];
+                buffer[i++] = pixel.R;
+                buffer[i++] = pixel.G;
+                buffer[i++] = pixel.B;
+                buffer[i++] = pixel.A;
+                if (!hasAlpha && pixel.A != 255)
+                    hasAlpha = true;
+            }
+        }
+        return buffer;
+    }
+}
--- a/src/Lumper.UI/ViewModels/Bsp/BspViewModel.Search.cs
+++ b/src/Lumper.UI/ViewModels/Bsp/BspViewModel.Search.cs
@@ -33,7 +33,7 @@ public partial class BspViewModel
     }
 
     private static async void Search(
-        (string?, MatcherBase?, BspNodeBase?) args)
+        (string?, MatcherBase?, BspNodeBase) args)
     {
         (string? pattern, var matcherBase, var bspNode) = args;
         if (matcherBase is null || pattern is null)
--- a/src/Lumper.UI/ViewModels/Bsp/Lumps/PakFile/PakFileEntryVtfViewModel.cs
+++ b/src/Lumper.UI/ViewModels/Bsp/Lumps/PakFile/PakFileEntryVtfViewModel.cs
@@ -1,11 +1,12 @@
 using System;
 using System.IO;
-using System.Runtime.InteropServices;
+using Lumper.Lib.BSP.Struct;
+using Lumper.UI.Models;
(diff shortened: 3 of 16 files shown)


tsa96 (changes requested):
Implementation looks solid, good job! I'm still wondering if we can centralize the VTF to image stuff to a single location so we don't have multiple parts of the UI doing the conversion, but maybe not that important.

Now I have a really annoying nit but it's just a refactor waiting to happen to I think best we do now: you're using "VTF" instead of "Vtf" all over the place. For BSPs we always use the Titlecase/UpperCamelCase naming (besides namespaces? idk why that is), let's keep naming consistent. I'll try get C# EditorConfig stuff in soon that screams at you about this.

Also, `VTFData` definitely shouldn't be directly in `Models`, should be something like `Models/Vtf/VtfFileData.cs`.

Sorry for all the nits, code-wise stuff looks okay.

I built the branch and played around with stuff a bunch, here's a bunch of further feature we'll want - if you want to do a few of these now go ahead, but lets get this branch in ASAP once BIRD reviews and additional stuff can come in future PRs:

- [ ] Clicking a texture should open a new window showing that texture at full size
- [ ] Picking the width/height of textures on the grid, with a top bar for setting that
- [ ] Searching/filtering
- [ ] Filtering out cubemaps (something like `^c-?\d+_-?\d+_-?\d+[\.hdr]{0,}\.vtf$`), checkbox (default) in topbar for it
- [ ] (Long term) Check both the filename and hash of each texture against a precomputed list of assets from Valve games, and show icons on the thumbnail of any texture that matches (orange icon for name, red for hash). Related to https://github.com/momentum-mod/lumper/issues/3

ddorab:
- [ X ] VTF -> Vtf
- [ X ] VTFData definitely shouldn't be directly in Models, should be something like Models/Vtf/VtfFileData.cs.

- [ X ] Clicking a texture should open a new window showing that texture at full size
- [ X ] Picking the width/height of textures on the grid, with a top bar for setting that
- [ X ] Searching/filtering
- [ X ] Filtering out cubemaps (something like ^c-?\d+_-?\d+_-?\d+[\.hdr]{0,}\.vtf$), checkbox (default) in topbar for

src/Lumper.UI/Views/VtfBrowser/VtfBrowserView.axaml.cs
```diff
@@ -0,0 +1,35 @@
...
+    private void Item_PointerPressed(object? sender, PointerPressedEventArgs e)
+    {
+        if (e.ClickCount == 2)
+        {
+            var selectedVtf = (VtfBrowserViewModel.Vtf)((Border)sender!).DataContext!;
+
+            var window = new VtfImageWindow();
+            window.DataContext = new VtfImageWindowViewModel { Image = selectedVtf.Image, Name = selectedVtf.Name};
+            window.Show();
+        }
```
tsa96, line 32: Does `DoubleTapped` not work here? From docs I can't tell if that's just for touchscreen or also works for cursors.

ddorab: seems to work yes. PointerPressed was from stackoverflow, I didn't read the docs on this one

tsa96 (approved):
Okay, lgtm but let's get BIRD's feedback. Might be worth posting an updated screenshot with all the new goodies that you showed me in voice.

ddorab:
sure
resim
resim
resim

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
```diff
@@ -0,0 +1,125 @@
...
+        private Image<Rgba32>? _image;
+        private string _name = "";
+
+        public Image<Rgba32>? Image
+        {
+            get => _image;
+            set => this.RaiseAndSetIfChanged(ref _image, value);
+        }
+
+        public string Name
```
BIRD311, line 27: nit: I like the private variable next to the public getter/setter but I know this isn't done very consistent in this project, but it is a few lines later so might as well mention it

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
```diff
@@ -0,0 +1,125 @@
...
+    {
+        get => _dimensions;
+        set
+        {
+            this.RaiseAndSetIfChanged(ref _dimensions, value);
+            this.RaisePropertyChanged(nameof(MaxNameWidth));
+        }
+    }
+
+    public uint MaxNameWidth => (uint)_dimensions - 8;
```
BIRD311, line 46: at least comment the magic number .. whats that for?

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
```diff
@@ -0,0 +1,125 @@
...
+        get => _showCubemaps;
+        set
+        {
+            this.RaiseAndSetIfChanged(ref _showCubemaps, value);
+            this.RaisePropertyChanged(nameof(TextureBrowserItems));
+            this.RaisePropertyChanged(nameof(TexturesCount));
+        }
+    }
+
+    [GeneratedRegex(@"^((c-?\d+_-?\d+_-?\d+)|cubemapdefault)[\.hdr]{0,}\.vtf$")]
```
BIRD311, line 61: that's also something I'd like to have a comment for .. they tend to be hard to read if its not split in its parts .. also `[\.hdr]` matches `.hdr.r...ddddd.hdhdhh` and thats not intended, is it?

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
```diff
@@ -0,0 +1,125 @@
...
+    private bool _showCubemaps;
+
+    public bool ShowCubemaps
+    {
+        get => _showCubemaps;
+        set
+        {
+            this.RaiseAndSetIfChanged(ref _showCubemaps, value);
+            this.RaisePropertyChanged(nameof(TextureBrowserItems));
+            this.RaisePropertyChanged(nameof(TexturesCount));
```
BIRD311, line 57: this should be called when the list changes and not when somehing which is goning to change the list is called, right? also, does the count in UI actually change? didn't see it

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    [GeneratedRegex(@"^((c-?\d+_-?\d+_-?\d+)|cubemapdefault)[\.hdr]{0,}\.vtf$")]
+    private static partial Regex _rgxCubemap();
+
+    private static ObservableCollection<Vtf> _textureBrowserItems =
```
BIRD311, line 64: don't like that this is static .. can it not live here?

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
```diff
@@ -0,0 +1,125 @@
...
+    private string _textureSearch = "";
+
+    public string TextureSearch
+    {
+        get => _textureSearch;
+        set
+        {
+            this.RaiseAndSetIfChanged(ref _textureSearch, value);
+            this.RaisePropertyChanged(nameof(TextureBrowserItems));
+            this.RaisePropertyChanged(nameof(TexturesCount));
```
BIRD311, line 101: this again

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
```diff
@@ -0,0 +1,125 @@
...
+        using var mem = new MemoryStream();
+        entry.DataStream.CopyTo(mem);
+        var vtfFileData = new VtfFileData(mem.ToArray());
+
+        _textureBrowserItems.Add(new Vtf
+        {
+            Name = name,
+            Image = vtfFileData.GetImage(0, 0, 0, 0)
+        });
+    }
```
BIRD311, line 119: I don't like this, it get called in the constructor and creates *all* the images even if the user never opens the viewer .. it also does the same as the other PakFile thing (as your comment says) .. haven't decided how we should do this, but its either using the same `Image<Rgba32>` as the other class (so using the `Vtf` class in both or keeping a reference of the PakFile class here and using its image) or at least doing some of the work here in a getter so it doesn't get called right away .. because you need all images for the view, it will still use a lot of memory for this and we should probably clear some of it but I guess we can improve the viewers performance later .. the VTFs have a thumbnail I was playing around with but its very low resolution

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
```diff
@@ -0,0 +1,125 @@
...
+
+    public static void ClearTextures()
+    {
+        _textureBrowserItems.Clear();
+    }
```
BIRD311, line 124: only needed because the list is static

src/Lumper.UI/ViewModels/VtfBrowser/VtfImageWindowViewModel.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        get => _name;
+        set => this.RaiseAndSetIfChanged(ref _name, value);
+    }
+
+    public Image<Rgba32>? Image
+    {
+        get => _image;
+        set => this.RaiseAndSetIfChanged(ref _image, value);
+    }
```
BIRD311, line 22: same as `class Vtf`?

src/Lumper.UI/ViewModels/Bsp/Lumps/PakFile/PakFileEntryVtfViewModel.cs
```diff
@@ -23,198 +25,112 @@ public class PakFileEntryVtfViewModel : PakFileEntryLeafViewModel
...
 
     private bool _isModified = false;
-    public override bool IsModified { get => _isModified; }
+
+    public override bool IsModified
+    {
+        get => _isModified;
+    }
 
     public Image<Rgba32>? _image = null;
```
BIRD311, line 42: should be private (my fault, but you've fixed the others)

BIRD311 (changes requested):

src/Lumper.UI/Models/Vtf/VtfFileData.cs
```diff
@@ -0,0 +1,90 @@
...
+using VTFLib;
+
+namespace Lumper.UI.Models;
+
+public class VtfFileData
+{
+    private bool _opened;
+    private readonly uint _imageIndex;
+
+    public readonly uint _frameCount;
```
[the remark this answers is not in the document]
ddorab: yeah so I'm not a fan of this either but it would've been too many lines otherwise so I just thought you or tom would say something, I'll change that

src/Lumper.UI/ViewModels/Bsp/Lumps/PakFile/PakFileEntryVtfViewModel.cs
```diff
@@ -15,6 +16,7 @@ public class PakFileEntryVtfViewModel : PakFileEntryLeafViewModel
         PakFileEntry entry, string name)
         : base(parent, entry, name)
     {
+        VtfBrowserViewModel.AddTexture(entry, name);
```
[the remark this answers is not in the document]
ddorab: I think it's best to move parsing stuff into the library but this seemed like the cleanest solution for now

src/Lumper.UI/ViewModels/Bsp/Lumps/PakFile/PakFileEntryVtfViewModel.cs
```diff
@@ -23,198 +25,112 @@ public class PakFileEntryVtfViewModel : PakFileEntryLeafViewModel
...
 
-        FrameCount = VTFFile.ImageGetFrameCount();
-        FaceCount = VTFFile.ImageGetFaceCount();
-        MipmapCount = VTFFile.ImageGetMipmapCount();
-        Flags = (VTFImageFlag)VTFFile.ImageGetFlags();
+        this.RaisePropertyChanged(nameof(FrameMax));
+        this.RaisePropertyChanged(nameof(FaceMax));
+        this.RaisePropertyChanged(nameof(MipmapMax));
 
         Info = $"MajorVersion: {VTFFile.ImageGetMajorVersion()}\n" +
```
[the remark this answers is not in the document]
ddorab: which line(s) are you talking about?

src/Lumper.UI/ViewModels/Bsp/Lumps/PakFile/PakFileEntryVtfViewModel.cs
```diff
@@ -224,49 +140,56 @@ public static Image<Rgba32> ImageFromFileStream(Stream fileSteam)
...
+
         _isModified = true;
-        VTFFile.BindImage(imageIndex);
         byte[] buffer = GetRgba888FromImage(image, out _);
 
         var f = VTFFile.ImageGetFormat();
         int size = (int)VTFFile.ImageComputeImageSize(
             (uint)image.Width, (uint)image.Height, 1, 1, f);
-        var buffer2 = new byte[size];
+        byte[] buffer2 = new byte[size];
```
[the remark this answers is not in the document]
ddorab: not sure why/when I did that :D

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
ddorab: will change

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
ddorab: well I thought the variable names would be enough but I guess without any context it's not that obvious, it's just to pad the vtf name by 8 pixels, tell me if it needs a comment

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
ddorab: resim
um? will add a comment though

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
ddorab: well `_showCubemaps` is used in `TextureBrowserItems` but I don't really get what you mean, and yes, the count does change
resim

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
ddorab: that moving stuff to the library thing again

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
ddorab: yeah we should discuss this on discord

src/Lumper.UI/ViewModels/VtfBrowser/VtfImageWindowViewModel.cs
ddorab: yeah, I wasn't quite sure what  to do

src/Lumper.UI/ViewModels/Bsp/Lumps/PakFile/PakFileEntryVtfViewModel.cs
ddorab: will do

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
tsa96: BIRD is right, this uses a character set when it should use a capturing group, probably my fault since I wrote the initial version of this regex.

`cubemapdefault.hrdrdrrr.vtf` will match this

ddorab: I see okay, `^((c-?\d+_-?\d+_-?\d+)|cubemapdefault)(\.hdr){0,}\.vtf$` should be fine then

src/Lumper.UI/ViewModels/Bsp/Lumps/PakFile/PakFileEntryVtfViewModel.cs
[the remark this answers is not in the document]
BIRD311: 116 so `Info` and all the `VTFFile.something()` calls

src/Lumper.UI/ViewModels/VtfBrowser/VtfBrowserViewModel.cs
BIRD311: its just a good idea in general to have comments or variables for magic numbers so people know what it is in the future

src/Lumper.UI/ViewModels/Bsp/BspViewModel.Search.cs
```diff
@@ -40,6 +40,6 @@ private void SearchInit()
             return;
         //TODO: Add lock when search is slower than throttle rate
         var matcher = matcherBase.ConstructMatcher(pattern.Trim());
-        await bspNode.Filter(matcher);
+        await bspNode!.Filter(matcher);
```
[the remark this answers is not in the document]
ddorab: yeah because it's the `BspNode` from the `BspViewModel`, am I missing something?

ddorab:
@BIRD311 @tsa96 I've pushed commit that's addressing all the things I've resolved, does it all look good? As for the ones remaining, I need a bit of guidance

PR #40, opened 2024-06-24
Update readme

tsa96:
As discussed, separate PR with new readme so we can get the repo updated before my main PR is merged.

If someone could approved ASAP that'd be great, you can see it in action by looking at the main page of the repo with this branch selected.


README.md
```diff
@@ -1,8 +1,101 @@
...
+- __Stripper__ - Apply [Stripper:Source](https://www.bailopan.net/stripper/) configs to a BSP. What is usually done at
+  runtime on a server via SourceMod can be *permanently* applied to the BSP file!
+- __Replace Textures__ - Update the texdata lumps to refer to different material files. Supports regex!
+- __Run External Tool__ - Runs a given executable on the loaded BSP.
+
+![Jobs](images/jobs.png)
+
+### Compressed Exporting
+
+Lumper can load and save LZMA compressed BSPs, including the pakfile, producing the same result as `bspzip`.
```
Gocnak, line 55: Should we link to VDC on bspzip? https://developer.valvesoftware.com/wiki/BSPZIP

Gocnak (changes requested):

tsa96:
Cheers, done all those

README.md
```diff
@@ -1,8 +1,103 @@
...
+It is *not* a map authoring tool like Hammer, which operates on map source files (VMFs) that are then
+compiled to BSPs. Lumper is intended for small, destructive tweaks to existing maps. It's most similar to
+[VIDE](http://www.riintouge.com/VIDE/), which is unfortunately closed-source, inactive, and doesn't support more recent
+BSP versions.
+
+## Features
+
+### Entity Editor
+
+View and edit the contents of the editor lump à la EntSpy/VIDE. Includes a "Raw Entities" view for editing the entire
```
Gocnak, line 22:
```suggestion
View and edit the contents of the entity lump à la EntSpy/VIDE. Includes a "Raw Entities" view for editing the entire

```

README.md
```diff
@@ -1,8 +1,103 @@
...
+lump as plaintext.
+
+![Entity Editor](images/entity_editor.png)
+
+### Pakfile Editor
+
+View and edit the contents of the pakfile lump with a built-in file tree, text editor, and VTF viewer.
+It supports import files/directories, exporting to the filesystem, and can automatically update references to paths in
```
Gocnak, line 30:
```suggestion
It supports importing files/directories, exporting to the filesystem, and can automatically update references to file paths in

```

Gocnak (changes requested):

tsa96:
Jesus christ what is wrong with me

Gocnak (approved):

README.md
```diff
@@ -1,8 +1,103 @@
 ![Momentum Mod](https://momentum-mod.org/assets/images/logo.svg)
 
-> *Momentum Mod is a standalone game built on the Source Engine, aiming to centralize movement gametypes found in CS:S, CS:GO, and TF2.*
+> *Momentum Mod is a standalone game built on the Source Engine, aiming to centralize movement gametypes found in CS:S,
+CS:GO, and TF2.*
```
BIRD311, line 4: 120 char line length?

README.md
```diff
@@ -1,8 +1,103 @@
...
-> *Momentum Mod is a standalone game built on the Source Engine, aiming to centralize movement gametypes found in CS:S, CS:GO, and TF2.*
+> *Momentum Mod is a standalone game built on the Source Engine, aiming to centralize movement gametypes found in CS:S,
+CS:GO, and TF2.*
 
-Lumper is a .NET based tool developed by the Momentum Mod Team for manipulating BSP files, the Source engine's compiled map format.
-It is aimed at map reviewers and porters, giving them the ability to analyse and tweak BSP files without having to decompile then recompile.
+<h1 style="display: flex; align-items: center; gap: 0.5rem;">
+    <img src="images/Lumper.svg" alt="Lumper Logo" width="32" height="32">
+	<span>Lumper</span>
+</h1>
```
BIRD311, line 9: not markdown but I didn't find a good solution right now either :/

README.md
```diff
@@ -1,8 +1,103 @@
...
+- View the important parts of the BSP for moderation during map submission, *especially* the pakfile lump.
+- Remove packed assets we shouldn't be storing/serving (usually Valve stuff, e.g. CS:S textures packed for use in
+  CS:GO).
+- Tweak entities to better use native Momentum entities, and remove unnecessary ones, often via Stripper configs.
+- Move pesky music files to the `sound/music/` directory of the paklump so they're affected by the right volume sliders
+  ingame.
+
+## Technical Details
+
+Lumper is written in C# 12, targeting .NET 8.0. It's divided into three subprojects:
```
BIRD311, line 84: Don't think this adds much. Its pretty obvious and the versions are easy to find and having this here is just a chore to update.

README.md
```diff
@@ -1,8 +1,103 @@
...
+as [`bspzip`](https://developer.valvesoftware.com/wiki/BSPZIP).
+
+### JSON Diffs
+
+Lumper can serialize the contents of the BSP to JSON files, helpful for analysing a BSP in a human-readable format and
+diffing BSP versions in Git.
+
+## Download
+
+See the [Releases](https://github.com/momentum-mod/lumper/releases) tab on the right! This program is very new, and we
```
BIRD311, line 65: "new" .. I wonder who is going to change this and when. :) I would get rid of the second sentence here. People don't expect everything they download from github to auto-update, do they?

tsa96: I agree about the dotnet version in the readme, but in this case I think it's fine. We're likely to change release stuff in the future and I don't think it's at all unreasonable to update the readme when the program is more stable.

README.md
tsa96: Yes, that's what I currently have configured, though I just noticed I wasn't setting `max_line_length` in our editorconfig (Roslyn doesn't use it so doesn't matter to us very much).

Are you against this value in particular? This works fine on Github, not sure what your point is!

BIRD311: 80 is the common one and the editor config in this commit was 100 and I just wanted to mention it because its a third number. I like 80 but whatever, no point arguing about this.

tsa96: I'm not trying to argue about this. Commenting "120 char line length?" just doesn't clearly convey what your point is.

Decided on formatting rules we like is worth trying to get out the way with now since this PR reformats everything, I'd you prefer we go down to 80 that's completely fine. Roslyn unfortunately [doesn't enforce a maximum line length](https://github.com/dotnet/roslyn/issues/47158) but we could just use 80 everywhere, could also including a Resharper formatting config for anyone using it.

I just set my max line length in Rider/Resharper to 80 and auto-formatted, was surprised by how good it looked, also since we use top-level namespaces now, we shave 4 spaces off most lines. That being said, I worry that without a strict autoformatter like Prettier, wrapping at 80 is likely to provoke more formatting disagreements in certain cases.

I'm not overly attached to any value, so long as we pick one and use it consistently. I'm most inclined to stick with 120 since it probably saves some formatting nitpicking, and saves me having to go through every file checking formatting again and fixing comment wrapping, but I've no problem doing that if you're prefer 80/100. Maybe best to discuss with others in Discord as well?
?

tsa96: Also, if we were to use 80 we'd probably only want to enforce it in C# files, and not XAML. Some files like EntityEditorView.axaml become pretty unreadable, and I don't like breaking stuff up into separate controls just for the sake of lowering indentation since I find it makes layouting a lot more annoying, especially when using the designer view.

BIRD311: was more a preemptive "I don't want to argue" because the number always seems so arbitrary. I like 80 because it lets me fit stuff better on smaller screens and side by side on bigger ones. Obviously depends on fontsize as well and is just personal preference.

would still be nice to have autoformating so formatting nitpicks just aren't a thing

and yes, line length should be enforced for XAML or at least not 80 chars

tsa96: Pinged you on Discord about this a couple days ago, lets finish discussing formatting stuff there / in meeting today.

Issue #42, opened 2024-06-26
Automatic Updates


Gocnak:
Lumper should be able to detect a new version on GitHub and link to downloading the new release for users. The version could be parsed via the source files on GitHub, or we can use the github API to check the releases page for latest and see if the tag version matches the currently deployed one.

Long term it would be great for Lumper to auto download the new version and restart using it.

tsa96:
Yeah, just pinging github API should work fine.

Probably most appropriate in a dedicated service, like BspService and PageService. You can access the Main Window via Program.Desktop and display a MsBox.Avalonia message box if out of date.

I started using HttpClient in my commit yesterday, be sure to use that, not the old WebClient thing.

tsa96:
Assuming we're not actually any kind of installer and just linking the page, this is a relatively small issue, only real work is figuring out the github API endpoint, probably like 100 LOC tops.

tsa96:
On mobile which doesn't seem to support adding new labels, but this is a Good First Issue!

Gocnak:
This was done in #78 right?

PR #50, opened 2024-07-16
Auto-update functionality

dbampersand:
This is regarding this issue: https://github.com/momentum-mod/lumper/issues/42

The code in this PR first checks if the version number defined in <AssemblyVersion> does not match the most recent version on Github. If it doesn't, it downloads the zip, unpacks it and then copies it over the root directory. My linux partition is broken right now so only the Windows sections of the code have been tested.

Let me know if there's anything I can improve on in this PR!

Thanks!


tsa96:
Really nice work! This approach seems reasonable though I'm a little confused as to how it overwrites the initial assembly after exiting - don't you need to schedule the shell command doing this for after the program exits?

One big issue though, we have a PR open on `refactor/cleanup` that drastically reworks the UI codebase, as well as introducing quite a lot of linter rules. It's taking a while to review since it's a massive PR, but I'm not too keen on merging this into the main branch since we'd then have to do a bunch more commits on `refactor/cleanup` to get this in order.

I think the easiest approach would be to rebase your branch onto `refactor/cleanup`, and then follow all the linter rules it brings up, saves me doing it in code review. Since this PR is mostly adding a new class, doing the rebase should be quite easy, though happy to help if needed. Depending on how review of `refactor/cleanup` goes, we may just merge this branch into `refactor/cleanup` (before `refactor/cleanup` is merged into `main`), or just leave this PR unmerged but finished until we've merged `refactor/cleanup` in `main` first.

Once you've rebased onto `refactor/cleanup`, run `dotnet format analzyers` to run [dotnet format](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-format), fixing what it can and pointing warnings for any other issues. Most editors with C# language server support should also highlight these for you. I'll go through and review now and try to point out the things that come up anyway, but keep an eye out for any other warnings, they'll all need to be fixed.

Sorry this is all so chaotic. Once we have my humongous PR is, all this linter and formatting stuff will be standardised and code review should become a lot easier.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+
+                if (int.TryParse(currentVersion[0].ToString(), out version.major)
+                    && int.TryParse(currentVersion[1].ToString(), out version.minor)
+                    && int.TryParse(currentVersion[2].ToString(), out version.patch))
+                {
+                }
+                else
+                {
+                    throw new Exception("Could not parse Major/Minor/Patch version.");
+                }
```
tsa96, line 72: You've already verified that this regex fits this pattern so the TryParse should always be good - if the pattern was bad we'd have just returned a version `0.0.0`. So I would just use `int.Parse`.

I realise I've completely torn this method apart, I'm just a regex nerd and spent some time learning C# regexes lately (C# regex support is [*really* good](https://www.youtube.com/watch?v=ptKjWPC7pqw)!)

Also, I'd make this a static method on `Version` (previously MMP) just called `Parse`
Here's how I'd write the whole thing, though haven't had time to test and I might be getting the capture groups stuff slightly wrong:
```cs
private static Version Parse(string s)
{
  Match match = VersionRegex.Match(s);

  if (!match.Success || match,Captures.Count != 3)
    throw new InvalidDataException($"Failed to parse version for string {s}");

  return new MMP(
    int.Parse(match.Captures[0].Value),
    int.Parse(match.Captures[1].Value)
    int.Parse(match.Captures[2].Value)
  )
}

[GeneratedRegex(@"^(\d+)\.(\d+)\.(\d+)$")]
private static partial Regex VersionRegex();
```

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+        {
+            int exitCode;
+            ProcessStartInfo processInfo = new ProcessStartInfo();
+            Process process;
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                processInfo = new ProcessStartInfo("cmd.exe", "/c " + command);
+                processInfo.CreateNoWindow = true;
+                processInfo.UseShellExecute = false;
```
tsa96, line 88: Use object initializers here and below, e.g. 
```cs
processInfo = new ProcessStartInfo("cmd.exe", "/c " + command) { CreateNoWindow = true, UseShellExecute = false };
```

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+        {
+            HttpClient client = new HttpClient();
+
+            var request = new HttpRequestMessage()
+            {
+                RequestUri = new Uri("https://api.github.com/repos/momentum-mod/lumper/releases"),
+                Method = HttpMethod.Get
+            };
+            client.DefaultRequestHeaders.Add("User-Agent", "Other");
+            HttpResponseMessage m = client.Send(request);
```
tsa96, line 113: Use SendAsync

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            if (m.IsSuccessStatusCode)
+            {
+                string response = await m.Content.ReadAsStringAsync();
+
+                var assets = JsonConvert.DeserializeObject<GHUpdate>(JsonObject.Parse(response)[0].ToString());
+                return assets;
```
tsa96, line 120: Unncessary inline var, just do `return JsonConvert.DeserializeObject ... `

src/Lumper.UI/Updater/Updater.cs
tsa96, line 113: Weird variable name - I'd use "response"

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+
+
+        }
+        //Major/Minor/Patch format
+        public class MMP
+        {
+            public int major;
+            public int minor;
+            public int patch;
+            public MMP(int Major, int Minor, int Patch)
```
tsa96, line 46: Bit of a nit but I think it's clearer just calling this `Version`.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+                throw new Exception("Could not connect - error " + m.StatusCode);
+        }
+        
+        public static async Task<bool> CheckForUpdate()
+        {
+            GHUpdate assets = await GetGithubUpdates();
+
+            //parse tag name to find the current and latest version
+            //finding the format of xx.yy.zz
+            string newestVersionSplit = Regex.Match(assets.tag_name, "[0-9]+\\.[0-9]+\\.[0-9]+").ToString();
```
tsa96, line 132: Okay, let's not do the `^` and `$` around the GeneratedRegex, then we can reuse it here though. Capture group stuff should mean they're not needed in `Version.Parse`

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+                processInfo.CreateNoWindow = true;
+                processInfo.UseShellExecute = true;
+            }
+            processInfo.RedirectStandardError = true;
+            processInfo.RedirectStandardOutput = true;
+            processInfo.WorkingDirectory = System.IO.Directory.GetCurrentDirectory();
+            process = Process.Start(processInfo);
+        }
+
+        //Grab the JSON from the github API and deserialize
```
tsa96, line 102: Use `/// <summary>` tags (they're silly but whatever it's C# convention)

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+using Avalonia.Controls;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System.IO;
+
+namespace Lumper.UI.Updater
+{
+    internal class Updater
+    {
+        //struct for deserializing JSON objects
+        private struct Asset
```
tsa96, line 23: Use `/// <summary>` tags, also I'd say "For deserializing Github API response data" or something

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+        {
+            GHUpdate assets = await GetGithubUpdates();
+
+            //parse tag name to find the current and latest version
+            //finding the format of xx.yy.zz
+            string newestVersionSplit = Regex.Match(assets.tag_name, "[0-9]+\\.[0-9]+\\.[0-9]+").ToString();
+            MMP current = GetMMPVersion(Assembly.GetExecutingAssembly().GetName().Version.ToString());
+            MMP latest = GetMMPVersion(newestVersionSplit);
+
+            return (current.major != latest.major || current.minor != latest.minor || current.patch != latest.patch);
```
tsa96, line 136: Once you use `record` you can just do `current != latest`

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+        //struct for deserializing JSON objects
+        private struct Asset
+        {
+            [JsonProperty("browser_download_url")]
+            public string browser_download_url;
+            [JsonProperty("name")]
+            public string name;
+        }
+        //struct for deserializing JSON objects
+        private struct GHUpdate
```
tsa96, line 31: Would prefer `GithubUpdate`

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+            //parse tag name to find the current and latest version
+            //finding the format of xx.yy.zz
+            string newestVersionSplit = Regex.Match(assets.tag_name, "[0-9]+\\.[0-9]+\\.[0-9]+").ToString();
+            MMP current = GetMMPVersion(Assembly.GetExecutingAssembly().GetName().Version.ToString());
+            MMP latest = GetMMPVersion(newestVersionSplit);
+
+            return (current.major != latest.major || current.minor != latest.minor || current.patch != latest.patch);
+        }
+        //returns the URL to the download link for the OS-specific version
+        private static string GetPath(GHUpdate assets, string OS)
```
tsa96, line 139: Call this githubUpdate, not assets
Also method name is very vague, "GetDownloadUrlForOS"? Then I wouldn't even bother with a docs comment above.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+            for (int i = 0; i < assets.assets.Length; i++)
+            {
+                if (assets.assets[i].name.ToLower().Contains(OS.ToLower()))
+                {
+                    return assets.assets[i].browser_download_url;
+                }
+
+            }
+            return null;
+        }
```
tsa96, line 150: Use a `foreach` loop, also can use a StringComparison overload:
```cs
foreach (Asset asset in githubUpdate.assets)
    if (asset.Name.Contains(OS, StringComparison.OrdinalgnoreCase)
        return asset.BrowserDownloadUrl;

return null;
```

Note that method method can return `null`, so needs to be annotated that what, then you should check for that in below code.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            string newestVersionSplit = Regex.Match(assets.tag_name, "[0-9]+\\.[0-9]+\\.[0-9]+").ToString();
+            MMP latest = GetMMPVersion(newestVersionSplit);
+
+
+            //NOTE: linux is untested
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                string fileURL = GetPath(assets, "linux");
+                string fileName = "linux_" + latest.major + "." + latest.minor + "." + latest.patch + ".zip";
```
tsa96, line 165: Use interpolation, `string fileName = $"linux_{latest.major} ...`

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,215 @@
...
+                }
+
+            }
+            return null;
+        }
+
+        public static async ValueTask Update()
+        {
+
+            GHUpdate assets = await GetGithubUpdates();
```
tsa96, line 155: Once you get this rebased onto `refactor/cleanup`, there's a class called `IoHandler` and `IoProgressWindow` which you can use to download over HTTP. See the new code for downloading BSPs over HTTP for example of usage here. Would look really like to show progress and we download, extract, and restart!

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,215 @@
...
+            //NOTE: linux is untested
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                string fileURL = GetPath(assets, "linux");
+                string fileName = "linux_" + latest.major + "." + latest.minor + "." + latest.patch + ".zip";
+                string directoryName = fileName + "temp";
+
+                //download and unzip to a temp directory
+                WebClient webClient = new WebClient();
+                webClient.DownloadFile(new Uri(fileURL), fileName);
```
tsa96, line 170: Use HttpClient not WebClient, they both do same stuff, HttpClient is just newer.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,215 @@
...
+
+                if (Directory.Exists(directoryName))
+                {
+                    Directory.Delete(directoryName, true);
+                }
+
+                System.IO.Compression.ZipFile.ExtractToDirectory(fileName, directoryName);
+
+                //copy files from the temp directory to the root directory, then delete the temp directory and run the program again
+                ExecuteCommand("yes | cp -rf  \"" + System.IO.Directory.GetCurrentDirectory() + "/" + fileName + "temp\"" + "&& rm \"" + fileName + "\" && rm -rf \"" + System.IO.Directory.GetCurrentDirectory() + "/" + directoryName + "\"" + " && ./Lumper.UI");
```
tsa96, line 180: This is very hard to read. Let's use string interpolation (`$` at the start) plus (`@` at the start), pretty sure you can do both.
Also worth splitting each command into multiple lines.

src/Lumper.UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -34,7 +34,32 @@ public MainWindowViewModel()
...
+                .ShowWindowDialogAsync(Program.Desktop.MainWindow);
+
+            if (result != ButtonResult.Ok)
+                return;
+        }
+        else
+        {
+            ButtonResult result = await MessageBoxManager
+                .GetMessageBoxStandard(
+                    "No updates available.",
```
tsa96, line 55: Remove full stop

src/Lumper.UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -34,7 +34,32 @@ public MainWindowViewModel()
                 RxApp.MainThreadScheduler);
         }
     }
+    public async ValueTask UpdateCommand()
+    {
+        bool updateAvailable = await Lumper.UI.Updater.Updater.CheckForUpdate();
```
tsa96, line 39: Let's return the version, either by returning a nullable `Version` and checking for null, or keep the bool and use an out var. Will need to implement `ToString` on `Version` I guess.

src/Lumper.UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -34,7 +34,32 @@ public MainWindowViewModel()
...
+        if (updateAvailable)
+        {
+            ButtonResult result = await MessageBoxManager
+                .GetMessageBoxStandard(
+                    "Update Available",
+                    "Update available. Download now?", ButtonEnum.OkCancel)
```
tsa96, line 45: Let's be more verbose: "An update is available (version number), do you want to download and restart?

tsa96 (changes requested):

tsa96 (changes requested):

tsa96:
Oh and for commit messages please see https://github.com/momentum-mod#contribution-guidelines.
Sorry if Git stuff is a bit much, we just like the keep the tree as organised as possible and prefer to `fixup` changes brought up in review rather than separate commits. Let me know if you have any issues, and really appreciate you working through all these change requests!

dbampersand:
Hey - thanks a ton for the code review, it means a lot! I've rebased and made the requested changes.

There's significant code duplication in the method HttpDownload as I modified that from BspService.cs in order to grab the progress bar code - I think that could be replaced inside a more generic class but I don't want to step on toes while you're refactoring the UI. 

Let me know if this works.

Thanks!

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+                "Error",
+                "Could not parse version number.", ButtonEnum.Ok)
+            .ShowWindowDialogAsync(Program.Desktop.MainWindow);
+            return null;
+
+        }
+        if (current != latest)
+            return latest;
+        else
+            return null;
```
tsa96, line 175: These `Version`s are only ever going to be null if they *throw*, and if they do, we don't have to catch that here.

Like I said in the comments above, exceptions completely halt the execution of the current method, and jumps immediately to the first `catch` in the call stack. Since we're fine with catching all of these exceptions in one place (in `Update() somewhere, I'll point out below), it's fine to throw anywhere in these methods if something bad happens, then continue on in the below code with confidence that any condition that would have throw if true, is false. For example in the above regex stuff, when we do

```cs
if (!match.Success)
    throw new InvalidDataException("Could not parse version number");

GroupCollection currentVersion = match.Groups;

int.TryParse(currentVersion[1].ToString(), out var major);
 ... etc ...
 ```
`match.Success` is *guaranteed* to be true from the `GroupCollection currentVersion ... etc` line onwards.

So in this block, all we need is the following:
```cs
Version current = GetVersionFromString(Assembly.GetExecutingAssembly().GetName().Version.ToString());;
Version latest = GetVersionFromString(assets.TagName);

if (current != latest)
    return latest;
else
    return null;
```

An upshot of this approach is that return null now simply means that we have an up-to-date version, which is a completely standard thing we expect to happen all the time. If we returned null for parsing errors, we no longer know whether a return value of null is the result of regular, well-behaved behaviour (e.g. program is already up-to-date), or the result of an unintended parsing error (e.g. Github changed their API or something).

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+    }
+    public static async Task<Stream?> HttpDownload(string url, string fileName)
+    {
+        IoProgressWindow? progressWindow = null;
+        var buffer = System.Buffers.ArrayPool<byte>.Shared.Rent(80 * 1024);
+        var cts = new CancellationTokenSource();
+        var handler = new IoHandler(cts);
+
```
tsa96, line 183: We're doing other work after downloading, unzipping the file and moving it. So we should create the IoProgressWindow and related classes outside of this method, then pass them in. Then during this method push the progress bar along to 80% or so, then do the final 20% during the unzipping.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+                    handler.UpdateProgress(prog, $"{float.Floor((1 - ((float)remaining / length)) * 100)}%");
+                    await stream.WriteAsync(buffer.AsMemory(0, read), cts.Token);
+                    remaining -= read;
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            if (ex is TaskCanceledException)
+                LogManager.GetCurrentClassLogger().Info("Download cancelled by user");
```
tsa96, line 231: We should have a static logger on this class, our standard pattern is:
```cs
private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
```

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+    private static string GetPath(GithubUpdate assets, OSPlatform OS)
+    {
+        var operatingSystem = "";
+
+        if (OS == OSPlatform.Windows)
+            operatingSystem = "win";
+        else if (OS == OSPlatform.Linux)
+            operatingSystem = "linux";
+        else
+            throw new Exception("Unsupported OS: " + nameof(OS));
```
tsa96, line 262: `nameof(OS)` here is just "OS", just `OS` is fine.

Also using a non-specific exception is a linter warning, use something like `ApplicationException`

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+
+
+        }
+        catch (Exception ex)
+        {
+            ButtonResult result = await MessageBoxManager
+            .GetMessageBoxStandard(
+                "Error",
+                ex.Message, ButtonEnum.Ok)
+            .ShowWindowDialogAsync(Program.Desktop.MainWindow);
```
tsa96, line 298: We don't tend to use message boxes for showing errors since we have a built-in logger UI. The standard pattern we use is to create an NLog logger for any class that needs one with 

```cs
private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
```
then for erroring you'd use `Logger.Error(<your message>)`.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+        }
+        catch (Exception ex)
+        {
+            ButtonResult result = await MessageBoxManager
+            .GetMessageBoxStandard(
+                "Error",
+                "Could not parse version number.", ButtonEnum.Ok)
+            .ShowWindowDialogAsync(Program.Desktop.MainWindow);
+            return;
+        }
```
tsa96, line 315: This can all be in one big `try/catch`, then just do `Logger.Error(ex, "Failed to download update")` (in NLog you should put the error object as the first arg)

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+                Environment.Exit(0);
+            }
+            catch (Exception ex)
+            {
+                ButtonResult result = await MessageBoxManager
+                .GetMessageBoxStandard(
+                    "Error",
+                    "IO error: " + ex.Message, ButtonEnum.Ok)
+                .ShowWindowDialogAsync(Program.Desktop.MainWindow);
+                return;
```
tsa96, line 362: More message boxes that should be logger errors

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+        }
+
+
+        //NOTE: linux is untested
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+        {
+            try
+            {
+                var fileURL = GetPath(assets, OSPlatform.Linux);
+                var fileName = "linux_" + latest.ToString() + ".zip";
```
tsa96, line 324: if you do `$"linux_{latest}.zip"` tostring should get called for you!

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+            }
+            catch (Exception ex)
+            {
+                ButtonResult result = await MessageBoxManager
+                .GetMessageBoxStandard(
+                    "Error",
+                    "IO error: " + ex.Message, ButtonEnum.Ok)
+                .ShowWindowDialogAsync(Program.Desktop.MainWindow);
+                return;
+            }
```
tsa96, line 413: Message box grr

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+    private static Version? GetVersionFromString(string s)
+    {
+        //match pattern of xx.yy.zz
+        Match match = VersionRegex().Match(s);
+        if (match.Success)
+        {
+            GroupCollection currentVersion = match.Groups;
+
+
+            int.TryParse(currentVersion[1].ToString(), out var major);
```
tsa96, line 73: `TryParse` here returns a boolean whether the parse was successful, which you're not using. `TryParse` has an evil sibling, `Parse`, which will throw if it fails. But as I've rambled about below, throwing for a parsing error like that is completely fine. In fact it'd be *very* weird if the parse failed here, given our regex was successful! So all you need is 

```cs
return new Version(int.Parse(currentVersion[1].Value, currentVersion[2].Value, currentVersion[3].Value))`
```
note I'm using `.Value` on the GroupCollection instead of ToString. If you peek at the source, `ToString` just returns `Value`, but `Value` is already a string and is what you should be accessing.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Lumper.Lib.BSP.IO;
+using Lumper.UI.Views;
+using MsBox.Avalonia;
+using MsBox.Avalonia.Enums;
+using Newtonsoft.Json;
+using NLog;
+
+internal sealed partial class Updater
```
tsa96, line 19: Don't really think `internal` is necessary here.

tsa96 (changes requested):

tsa96:
> Hey - thanks a ton for the code review, it means a lot! I've rebased and made the requested changes.
> 
> There's significant code duplication in the method HttpDownload as I modified that from BspService.cs in order to grab the progress bar code - I think that could be replaced inside a more generic class but I don't want to step on toes while you're refactoring the UI.
> 
> Let me know if this works.
> 
> Thanks!

Getting there! Sorry for chucking so much at you, hope the comments are informative. If you have questions about anything or what to discuss I'm happy to talk about this in our Discord in text chat or voice.

I don't think the code duplication of HttpDownload is too bad, since I think your version should be a little different anyway, with the outer method controlling the IoHandler and IoProgressWindow.

src/Lumper.UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1,92 +1,194 @@
...
+            if (result != ButtonResult.Ok)
+                return;
+        }
+        else
+        {
+            ButtonResult result = await MessageBoxManager
+                .GetMessageBoxStandard(
+                    "No updates available",
+                    "No updates available", ButtonEnum.Ok)
+                .ShowWindowDialogAsync(Program.Desktop.MainWindow);
```
tsa96, line 61: Better to do a Logger.Info here I think. Maybe I'm weird but I don't like random message boxes popping up!

src/Lumper.UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1,92 +1,194 @@
...
+    {
+        Lumper.UI.Updater.Updater.Version? updateAvailable = await Lumper.UI.Updater.Updater.CheckForUpdate();
+        if (updateAvailable != null)
+        {
+            string updateNumber = updateAvailable.ToString();
+            ButtonResult result = await MessageBoxManager
+                .GetMessageBoxStandard(
+                    "Update Available",
+                    $"An update is available ({updateNumber}), do you want to download and restart?", ButtonEnum.OkCancel)
+                .ShowWindowDialogAsync(Program.Desktop.MainWindow);
```
tsa96, line 50: Actually sorry, I suggested returning null here but I think a bool would be *way* easier to understand. So make method signature would be 
`bool CheckForUpdate(out Version version)`

then in the method have something like
```cs
version = latest;
return (current != latest);
```

then here would be
```cs
        var updateAvailable = await Lumper.UI.Updater.Updater.CheckForUpdate(out Version version);
        if (updateAvailable)
        {
             ButtonResult result = await MessageBoxManager
                .GetMessageBoxStandard(
                    "Update Available",
                    $"An update is available ({version), do you want to download and restart?", ButtonEnum.OkCancel)
                .ShowWindowDialogAsync(Program.Desktop.MainWindow);
```

Also import is quite weird, just do `import Lumper.UI.Updater` then you can do just e.g. `Updater.CheckForUpdate()`

src/Lumper.UI/Views/MainWindow.axaml
```diff
@@ -1,57 +1,164 @@
...
+                        <TextBlock>Lumper is likely to contain bugs in its current state, so we highly recommend backing up your original BSP.</TextBlock>
+                      </StackPanel>
+                    </ToolTip.Tip>
+                  </materialIcons:MaterialIcon>
+                </StackPanel>
+              </CheckBox>
+              <Separator />
+              <MenuItem Header="_Export JSON data" Command="{Binding BspService.JsonDump}" />
+              <Separator />
+              <MenuItem Header="Check for updates" Command="{Binding UpdateCommand}" />
```
tsa96, line 75: Let's split this into an "About" section that just has this.

tsa96 (changes requested):

tsa96:
@BIRD311 do you want to take a look at this? Really hate to burden you with every more review work, but if you were up for testing on linux and maybe glancing at the code that'd be great. Up to you!

dbampersand:
Hey - thanks a ton for the code review and thanks for your patience. I've made the changes and squashed them. Hoping I did that correctly! The one change I couldn't do was regarding CheckForUpdate - C# doesn't support async with 'out' modifiers, so I changed that to a Tuple that contains the success/fail and the Version. I'll see if I can test the Linux code later.

tsa96:
On phone so hard to review but I noticed you're using Tuple when you can use the new tuple syntax of e.g. `(bool, Updater.Version)`, so return a `Task<(bool, Updater.Version)>`. Then you can save the Item1, Item2 crap by destructuring e.g. "`(var success, Updater.Version version) = await GetUpdate();` (I forget the rhs method name here)

For Git stuff, probably the easiest approach is to do a mixed reset to unstage everything, then just recommit in two different commits. To clarify what we want the commit tree to look like, we want

- a commit adding the assembly version: `feat(ui): specify AssemblyVersion of Lumper.UI`
- a commit adding all the updater code: `feat(ui): add auto updated system`

Anything brought up in review should just be fixup/amended to that second commit (just stage those changes then `git commit --amend --no-edit`)

I noticed dotnet format formatted a bunch of csproj files that it shouldn't, this is because we haven't finished the auto formatting system yet, please just discard any changes made to files you're not interacting with in this PR. For the assemblyversion thing, probably easiest to just discard all changes to that file then add the assemblyversion line back in your editor. Sorry this is so messy, will be sorted soon!

To recommit I'd do
```
git branch backup // make backup branch but don't checkout, in case you mess up you can reset back to here and start over , can delete at end
git reset --mixed HEAD~4 // or wherever the first commit before your first commit is
// remove csproj formatting stuff from -  lumper ui csproj
git add lumper.ui.csproj
git commit -m <above message I wrote>
git add updater.cs, MainWindowViewModel.cs, any other stuff you modified (try to avoid auto formatting stuff!)
git commit -m <other above message>
git push --force
```
Check logs to make sure yt makes sense before pushing. This is rushed sorry, on phone in a coffee shop and only have a few minutes. If struggling with any git stuff I highly recommend a graphical client, my preferred client is Fork https://git-fork.com/ (free, just slams asking for money occasionally)

dbampersand:
Thanks for the patience and the explanation - I've fixed the commit history now and it should be clean of the formatting edits.

tsa96:
Aaaalmost, just you've still got a fixup commit. If you just reorder the commit with the UI stuff and the commit with the assemblyversion stuff, you can easily amend any changes you have to make to the UI stuff using `git commit --amend --no-edit`, which will amend any staged changes into the most recent commit.

To order do `git rebase -i HEAD~3` and just swap the order of the commits in the text editor that Git will open, then save and close the editor, and you'll have switched the order around.

Also, I'd prefer you keep the commit messages shorter, as you can see in the commit messages when shown on Github get trimmed awkwardly. Also "updater" isn't really a valid scope, I'd just use the messages in my above message. You can rename during the interactive rebase above. When you enter the rebase, you'll see an editor looking something like this

```
pick aaaaaaaa feat(updater): added checking and automatically downloading/applying …
pick bbbbbbbb feat(Updater): added AssemblyVersion tag to Lumper.UI.csproj allowing…
pick cccccccc fixup! feat(updater): added checking and automatically downloading/ap…
```
to reorder and reword, change these lines to
```
reword bbbbbbbb feat(ui): specify AssemblyVersion of Lumper.UI
reword aaaaaaaa feat(ui): add auto updater system
fixup cccccccc fixup! feat(updater): added checking and automatically downloading/ap…
```
that'll reword the existing commits, switch their order, then fixup the commit with the "fixup!" on the front into the commit above it. Sorry if that's a bit overwhelming, but if you can make sense of that, you've just learnt probably the single most powerful feature in Git!

dbampersand:
Got it (I think!). Sorry, I'm not great with git but I appreciate the hand holding.

tsa96:
Yeah looks good. And no problem, we're pretty exacting with structuring commits so naturally we're happy to help people struggling with it. Git is a hassle to learn at first, with a steep learning curve, but given how ubiquitous it's become, it's well worth putting the time into learning!

I had a quick glance at your commits and mostly looking good though spotted a few small things, one that you're still using `Tuple<A, B>` over just `(A, B)`, mentioned in my commit above. It's really just a syntactic thing but would prefer using it. However, I'm exhausted and need to head off, so I'll go through stuff fully tomorrow morning,

Really appreciate the work by the way, and sticking through everything brought up in review. We'll soon be encouraging everyone working on map porting to start using Lumper, which is a vital part of getting Momentum in a releasable state. Great to have other devs working on it!

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,420 @@
...
+
+
+            int.TryParse(currentVersion[1].ToString(), out var major);
+            int.TryParse(currentVersion[2].ToString(), out var minor);
+            int.TryParse(currentVersion[3].ToString(), out var patch);
+
+            return new Version(major, minor, patch);
+        }
+        throw new Exception("Could not parse Major/Minor/Patch version.");
+        return null;
```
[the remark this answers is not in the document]
tsa96: You're still not throwing with `if (!match.Success) throw ...`, your current approach is a lot harder to read.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,406 @@
...
+    /// </summary>
+    /// <returns></returns>
+    public static async ValueTask Update()
+    {
+        GithubUpdate assets;
+        try
+        {
+            assets = await GetGithubUpdates();
+
+            Version? latest;
```
tsa96, line 262: Should be `Version latest = GetVersionFromString(assets.TagName);` note that latest is non-null. Idk if your editor has LSP supported but this should be warning at you about this, don't ignore it!

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,406 @@
...
+    private static string GetPath(GithubUpdate assets, OSPlatform OS)
+    {
+        var operatingSystem = "";
+
+        if (OS == OSPlatform.Windows)
+            operatingSystem = "win";
+        else if (OS == OSPlatform.Linux)
+            operatingSystem = "linux";
+        else
+            throw new ApplicationException("Unsupported OS: " + OS);
```
tsa96, line 234: We're picking this same string in `Update`, so let's just pass that in to this function. I'll post a rewritten version of `Update` below, you can modify this function to take a string input.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,406 @@
...
+        public int Major { get; } = major;
+        public int Minor { get; } = minor;
+        public int Patch { get; } = patch;
+        public override string ToString() => $"{major}.{minor}.{patch}";
+    }
+    /// <summary>
+    /// Parses a string to find the major/minor/patch versioning
+    /// </summary>
+    /// <param name="s">String containing a version number following the format 'xx.yy.zz'</param>
+    private static Version GetVersionFromString(string s)
```
tsa96, line 79: Better name is `ParseVersion`

tsa96:
Okay done, though haven't tested. Please take a look and make sure everything works, then fixup into your commit!

Spent a while trying to do nice JSON parsing but since we're using Newtonsoft stuff still, we can't use some of the nice new stuff outlined [here](https://www.stevejgordon.co.uk/sending-and-receiving-json-using-httpclient-with-system-net-http-json). Oh well!

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,406 @@
...
+                if (Directory.Exists(directoryName))
+                {
+                    Directory.Delete(directoryName, true);
+                }
+
+                float progress = DownloadProgressPercentage / 100.0f;
+
+                Directory.CreateDirectory(directoryName);
+                List<Task> tasks = new List<Task>();
+                List<FileStream> fileStreams = new List<FileStream>();
```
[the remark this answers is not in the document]
tsa96: Nvm these were totally unused!

dbampersand:
Thanks for that - I've commited a couple of fixes:

```
    private record Asset
    {
        [JsonProperty("browser_download_url")]
        public string BrowserDownloadUrl { get;}

        [JsonProperty("name")]
        public string Name { get; }
    }
```
```
    private record GithubUpdate
    {
        [JsonProperty("tag_name")]
        public string TagName { get; }

        [JsonProperty("assets")]
        public Asset[] Assets { get; }

    }
```
To work properly JsonConvert.DeserializeObject needs the setter otherwise it won't update the fields.

 ```               
    osName = "windows";
```

The file for Windows is along the lines of "Lumper_v0.1.1_win-x64.zip" so searching  if the string cotntains "Windows" fails.

**Comment by tsa96** [comment]
Oh gotcha, both those fixes make sense.

We're probably about ready to merge but going to give BIRD the opportunity to look at this, and not sure if we'll merge this into `refactor/cleanup` straight away, or wait until that branch is in first.

By the way, I have a big load of Lumper issues I was meaning to make issues for and haven't had a chance to, if you're interested in helping out further keep an eye on the issues board!

**Comment by tsa96** [comment]
Just thinking about this a little today, might be a better approach to pull the updating logic out to a separate executable so we can copy and replace stuff easier. Though I guess we'd have to do some work on startup of the main app to determine if we've just updated, and if so, replace the old version of the updater executable with the new one.

Not sure we need to do this, want to hear what others think in our team meeting tomorrow.

File: src/Lumper.UI/Updater/Updater.cs [hunk 39]
```diff
@@ -0,0 +1,362 @@
...
+                osName = "linux";
+            }
+            else
+            {
+                throw new ApplicationException("OS is not supported");
+            }
+
+            var downloadUrl = GetReleaseDownloadUrl(assets, osName);
+            var fileName = $"{osName}_{latest}.zip";
+            var directoryName = $"{fileName}temp";
```
BIRD311, line 286: write the download to `/tmp` so it will be cleaned up if something goes wrong and we don't make it to the `rm` in code. There should be a c# function to get the temp folder or a temp file independent of the platform we're on.

src/Lumper.UI/Updater/Updater.cs
```diff
@@ -0,0 +1,362 @@
...
+            // copying files from the temp directory to the root directory,
+            // then delete the temp directory and run the program again
+            var command =
+                os == OSPlatform.Linux
+                    ? $"""
+                           sleep 2 &&
+                           yes | cp -rf "{currentDirectory}\{directoryName}" &&
+                           rm "{currentDirectory}\{fileName}" &&
+                           rm -rf "{currentDirectory}\{directoryName}" &&
+                           ./Lumper.UI
```
BIRD311, line 340: - copy doesn't have a destination
- `rm -rf` is scary, try moving files instead
- you can't rely on the current directory being where `./Lumper.UI` is
- forward slashes in the path

BIRD311 (changes requested):
only skimmed this, don't personally care much about the feature. Thank for helping.

src/Lumper.UI/Updater/Updater.cs
tsa96: Wasn't sure how long this is guaranteed to last, turns out on Windows, the OS doesn't clean up the files for you! I was under the impression they were wiped on reboot, which AFAIK is what all unix-like systems with a `/tmp` do. So, we should still `rm` the file after.

tsa96:
Discussed this a bit yesterday and keen to avoid the shell scripts. Most of the code here we'll still use, but we'll probably pull some of it out to separate executable.

Going to wait until `refactor/cleanup` is in, then I'll start a new branch off there and pull in some of these changes and create a separate little project. That way I finish auto updating and release and QA CI jobs at once - expect I'll need the AssemblyVersion addition for that. So not sure if this PR itself will get merged in the future, but I'll make sure to cherry pick your current commits so you get credit.

tsa96:
Closing these since https://github.com/momentum-mod/lumper/pull/78 supersedes it.

Kept a lot of your code @dbampersand, mainly just changed how we handle the downloaded zip so we have to do fewer shell commands. Added you as a co-author on https://github.com/momentum-mod/lumper/commit/0fd201aedd8e80641562047b9cbc0c6f924be58f, really appreciate the contribution!

Issue #53, opened 2024-07-20
Brackets removed from VScript entity I/O after saving.


talascio:
In VScript-supported titles, the `RunScriptCode` entity input allows you to run any VScript functions directly in Entity I/O without needing to include a separate file.  Certain VScript functions (notably in TF2) accept a table as an argument, and when saving a modified BSP the brackets will be removed from the arguments, causing a script error.

The following entity I/O 

```
"OnSpawn" "beamprop2RunScriptCodeEmitSoundEx({sound_name = `Ambient.NucleusElectricity`, entity = self, volume = 0.2, flags = 1})0-1"
```
gets replaced with 

```
"OnSpawn" "beamprop2RunScriptCodeEmitSoundEx(sound_name = `Ambient.NucleusElectricity`, entity = self, volume = 0.2, flags = 1)0-1"
```

the `sound_name` variable doesn't exist outside of the scope of that table, so this will throw the error `the index 'sound_name' does not exist`.  When executed.

I have not seen how Entity I/O is parsed yet, however if it's split into an array/dict/whatever where you can check if the input is "RunScriptCode", it should preserve any brackets that may be in the parameter.

PR #54, opened 2024-07-21
Curly bracket removal / vscript values output fix

dbampersand:
Closes https://github.com/momentum-mod/lumper/issues/53

Curly brackets are currently being treated as special characters that only signify the start and end of sections, but they can appear in any entity - for instance, a valid entity name is {a} - this pr should fix this. (Regarding changes made to EntityLump.cs)

The second part of the change is regarding vscript outputs not saving correctly - previously, commas were used to delimit but when a file is compiled to use vscript, it instead uses \u001b but this was not being taken into account when writing the string to the disk, and as such was being loaded incorrectly (as a regular key/value pair rather than as a specific EntityIO) (Regarding changes made to EntityIO.cs)

Here's a bsp I made for TF2 to test this issue: https://filebin.net/5az4cvzf7mik93ys and it now loads and saves the values correctly (vscript is on the func_button, targetting 'test' with an input of RunScriptCode).

### DIFF
--- a/src/Lumper.Lib/BSP/Lumps/BspLumps/EntityLump.cs
+++ b/src/Lumper.Lib/BSP/Lumps/BspLumps/EntityLump.cs
@@ -55,6 +55,10 @@ public class EntityLump(BspFile parent) : ManagedLump<BspLumpType>(parent)
                             Data.Add(new Entity(keyValues) { EntityLumpVersion = Version });
                             return true;
                         }
+                        else
+                        {
+                            stringBuilder.Append(x);
+                        }
 
                         break;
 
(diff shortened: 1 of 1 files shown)


tsa96:
Nice job finding these.

Since the EntityIO stuff is currently in review I don't want to add this change on top of it, I've just applied the `ToString` fix on my commit that I'll force-push in a minute.

Please can you get rid of the changes to the EntityIo stuff, and just push the curly braces changes?

dbampersand:
Thanks - I've updated with just the curly brace changes.

tsa96 (approved):
Just tested this out and refreshed my memory of this parser, all looks good!

PR #56, opened 2024-07-21
Fix for parsing IO entries that have no named output

### LINKED ISSUE #55: File parsing skips IO entries that have no named output
When a IO node for an entity has no named 'output' it fails to load. Here's an example of non-functioning but valid IO: https://i.imgur.com/eryGDGm.png (the first column, 'My Output') and this isn't stripped when creating the bsp.

For an example bsp of this, take a look at c1a0c.bsp from Half Life Source: https://filebin.net/ojfr1evko74bjl6z

```Warn    EntityLump                Entity parser skipped value "hidden_bully_off,,,0,-1" for empty key, how u do dis?```

dbampersand:
Small fix for this issue: https://github.com/momentum-mod/lumper/issues/55
Closes #55

### DIFF
--- a/src/Lumper.Lib/BSP/Lumps/BspLumps/EntityLump.cs
+++ b/src/Lumper.Lib/BSP/Lumps/BspLumps/EntityLump.cs
@@ -36,7 +36,6 @@ public class EntityLump(BspFile parent) : ManagedLump<BspLumpType>(parent)
     {
         var stringBuilder = new StringBuilder(512);
         var keyValues = new List<KeyValuePair<string, string>>();
-
         try
         {
             string? key = null;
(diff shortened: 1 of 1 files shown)


dbampersand:
On reflection it might be better to just remove the key length check altogether? There are bsps that have null keys and a use case for Lumper could be in fixing them: https://i.imgur.com/34ZZeTK.png 

See c4a1e.bsp for an example of this: https://filebin.net/n6afdc9myzoyjgg9

tsa96 (approved):
Yeah, I think this approach is exactly right.

tsa96:
Reworded your commit to reflect what this change is actually doing now, allowing empty keys everywhere, not just for ent io.

Issue #60, opened 2024-07-24
Rename cubemaps when renaming map


tsa96:
When saving a map, if the pakfile contains VTF files matching `materials/maps/<current map name>/<cubemap regex>.vtf`, and the map name is changing, move those items to use the new map name.  We have a `CubemapRegex` already though I think it lives UI code atm, if so, move it to lib.

PR #61, opened 2024-07-25
Default cubemap renaming

dbampersand:
Fix for this issue: https://github.com/momentum-mod/lumper/issues/60

Didn't use ```CubemapRegex``` as matching cX_cY_cZ as well will cause false positives. (Unless I'm misunderstanding how that works)

### DIFF
--- a/src/Lumper.Lib/BSP/BspFile.cs
+++ b/src/Lumper.Lib/BSP/BspFile.cs
@@ -105,6 +105,15 @@ public sealed partial class BspFile : IDisposable
         if (path is null && FilePath is null)
             throw new ArgumentException("Not given a path to write to, and current BSP doesn't have a path");
 
+        if (path != null && Path.GetFileName(path) != FilePath)
+        {
+            PakfileLump pakfile = GetLump<PakfileLump>();
+
+            Dictionary<string, string> modified = pakfile.RenameCubemapsPath(Path.GetFileName(path));
+            foreach (KeyValuePair<string,string> modifiedPath in modified)
+                pakfile.UpdatePathReferences(modifiedPath.Value, modifiedPath.Key);
+        }
+
         string outPath;
         string? backupPath = null;
 
--- a/src/Lumper.Lib/BSP/BspFile.cs
+++ b/src/Lumper.Lib/BSP/BspFile.cs
@@ -111,7 +111,7 @@ public sealed partial class BspFile : IDisposable
 
             Dictionary<string, string> modified = pakfile.RenameCubemapsPath(Path.GetFileName(path));
             foreach (KeyValuePair<string,string> modifiedPath in modified)
-                pakfile.UpdatePathReferences(modifiedPath.Value, modifiedPath.Key);
+                pakfile.UpdatePathReferences(modifiedPath.Value, modifiedPath.Key, ".vmt");
         }
 
         string outPath;
--- a/src/Lumper.Lib/BSP/Lumps/BspLumps/PakFileLump.cs
+++ b/src/Lumper.Lib/BSP/Lumps/BspLumps/PakFileLump.cs
@@ -49,7 +49,7 @@ public partial class PakfileLump(BspFile parent) : ManagedLump<BspLumpType>(pare
     /// Updates all path references in the PakFileLump from oldPath to newPath,
     /// i.e. if a VMT references a VTF, it will be updated.
     /// </summary>
-    public void UpdatePathReferences(string newPath, string oldPath)
+    public void UpdatePathReferences(string newPath, string oldPath, string? limitExtension = null)
     {
         var opSplit = oldPath.Split('/');
         var npSplit = newPath.Split('/');
@@ -62,6 +62,8 @@ public partial class PakfileLump(BspFile parent) : ManagedLump<BspLumpType>(pare
 
         foreach (PakfileEntry entry in Entries)
         {
+            if (limitExtension is null || !entry.Key.EndsWith(limitExtension))
+                continue;
 
             var entryString = Encoding.Default.GetString(entry.GetReadOnlyStream().ToArray());
             var newString = entryString.Replace(oldPath, newPath, StringComparison.OrdinalIgnoreCase);
--- a/src/Lumper.Lib/BSP/Lumps/BspLumps/PakFileLump.cs
+++ b/src/Lumper.Lib/BSP/Lumps/BspLumps/PakFileLump.cs
@@ -105,11 +105,14 @@ public partial class PakfileLump(BspFile parent) : ManagedLump<BspLumpType>(pare
         string baseFilename = Path.GetFileNameWithoutExtension(newFileName);
         var entriesModified = new Dictionary<string, string>();
 
+        bool matched = false;
         foreach (PakfileEntry entry in Entries)
         {
             Match match = CubemapRegex().Match(entry.Key);
             if (match.Success)
             {
+                matched = true;
+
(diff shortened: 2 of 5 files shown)


tsa96:
On phone so not doing proper review but this should probably live in Lumper.Lib code so CLI can use it. Actually sounds like quite a helpful use for the cli, renaming BSP without breaking cubemaps (I have a WIP branch atm with CLI support for saving new BSPs)

dbampersand:
Great - I've moved it to be in lib and fixed the UI state not changing.

src/Lumper.UI/Services/BspService.cs
```diff
@@ -275,6 +278,12 @@ public async Task<bool> Save(IStorageFile? outFile)
             if (Program.Desktop.MainWindow is not null)
             {
                 var outName = outFile is not null ? Path.GetFileName(outFile.Path.AbsolutePath) : FileName;
+                if (outName != FileName)
+                {
+                    BspFile.GetLump<PakfileLump>().RenameDefaultCubemap(outName);
+                    BspService.Instance.PakfileLumpViewModel.IsModified = true;
+                    BspService.Instance.ResetLumpViewModels();
+                }
```
tsa96, line 286: Still don't think this is the right approach, should do this logic in the BspFile SaveToFile method. Probably worth adding a checkbox to the File... menu to toggle this in UI though, then pass that into SaveToFile.

I'd like to avoid reloading lump viewmodels here, since that would cause every UI page to reload. Unfortunately I can't think of a way to update the viewmodels that isn't ugly, but still worth it to avoid reloading the whole things. After SaveToFile is called, if successful, iterate through every PakfileEntryViewModel and update their Key to the Key of the underlying PakfileEntry, if changed.

Don't love doing this but should be perfectly performant, and will work if there's ever other pakfile changes we make during saving in the future.

src/Lumper.Lib/BSP/Lumps/BspLumps/PakFileLump.cs
```diff
@@ -38,6 +39,30 @@ public class PakfileLump(BspFile parent) : ManagedLump<BspLumpType>(parent), IFi
 
     private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
 
+    //as per https://github.com/ValveSoftware/source-sdk-2013/blob/master/mp/src/utils/vbsp/cubemap.cpp
```
tsa96, line 42: This doesn't seem right. Maps I'm looking at often have hundreds of cubemaps stored in `maps/<name>/c_<x coord>_<y coord>_<z coord>.vtf`, these should also be moved. grepping for "maps/" in that CPP file seems to find what I'm talking about: https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/utils/vbsp/cubemap.cpp#L512

tsa96 (changes requested):

dbampersand:
I've updated it for the logic to be in lib-only code and also fixed updating path references. Not confident on the way I've done updating the UI (the stuff in BspService), think I need to read more on how Avalonia actually works. I'll update later to clean up and I'll try to see if I can make the UI code less kludgy.

src/Lumper.Lib/BSP/BspFile.cs
```diff
@@ -105,6 +105,15 @@ public sealed partial class BspFile : IDisposable
         if (path is null && FilePath is null)
             throw new ArgumentException("Not given a path to write to, and current BSP doesn't have a path");
 
+        if (Path.GetFileName(path) != FilePath)
+        { 
+            Dictionary<string, string> modified = GetLump<PakfileLump>().RenameDefaultCubemap(Path.GetFileName(path));
+            foreach (KeyValuePair<string,string> modifiedPath in modified)
+            {
+                GetLump<PakfileLump>().UpdatePathReferences(modifiedPath.Value, modifiedPath.Key);
```
tsa96, line 113: Weird, UpdatePathReferences doesn't live in Lumper.Lib, I suspect you missed something from your commits.

src/Lumper.UI/Services/BspService.cs
```diff
@@ -290,6 +294,30 @@ public async Task<bool> Save(IStorageFile? outFile)
...
+                            PakfileLumpViewModel.DeleteEntry(pk);
+                }
+                // update model one more time
+                PakfileLumpViewModel.MarkAsModified();
+                PakfileLumpViewModel.UpdateEntries(true);
+                PakfileLumpViewModel.MarkAsModified();
+                PakfileLumpViewModel.UpdateModel();
+            }
+
+
```
tsa96, line 320: Shouldn't need to do this, just update keys on the viewmodel. The library code should just be updating keys on PakfileLump, which will cause the zip entries to get moved when the BSP is saved. Then you should be able to just loop through every pakfile entry viewmodel and update the key if changed. Should be all you need. Please ask on here or Discord if this stuff isn't make sense, I appreciate that the UI code can be hard to get your head around

tsa96 (changes requested):

dbampersand:
Oops, I totally did forget to stage that file. The UI update should be significantly simpler now.

src/Lumper.UI/Services/BspService.cs
```diff
@@ -11,9 +11,13 @@ namespace Lumper.UI.Services;
 using System.Threading;
 using System.Threading.Tasks;
 using Avalonia.Platform.Storage;
+using Avalonia.Rendering.Composition;
```
tsa96, line 14: ?

tsa96 (changes requested):
Sorry it took me a while to get to this, now that my big PR is in I'll try to be more responsive with code review. This'll need rebasing on latest `origin/main`.

Note that I'll be finally running an autoformatter this codebase soon that might make rebasing hard in the future, if that's the case and you're struggling to rebase, *please* just let me know and I'd be happy to resolve everything for you.

dbampersand:
Thanks for the review - I've rebased onto main and I've made those changes.

src/Lumper.Lib/BSP/Lumps/BspLumps/PakFileLump.cs
```diff
@@ -38,6 +41,91 @@ public class PakfileLump(BspFile parent) : ManagedLump<BspLumpType>(parent), IFi
...
+    //source engine only searches the following path: materials/maps/*FILENAME*/cubemapdefault(.vtf OR .hdr.vtf)
+    [GeneratedRegex(@"(materials)/(maps)/(.*)/((c-?\d+_-?\d+_-?\d+(.hdr){0,}.vtf)|(cubemapdefault(.hdr){0,}.vtf))")]
+    private static partial Regex DefaultCubemapRegex();
```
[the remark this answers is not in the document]
tsa96: Sorry, the one I provided before was wrong, with `d` instead of `\d`. I've cleaned up this regex a bit, maybe materials/maps/ bit optional.

tsa96 (approved):

tsa96:
Sorry I took long getting back to this, haven't had time for Lumper in a while.

I just resolved a couple things, mostly because the cubemap changes I were asking for in review didn't actually make sense, the VtfFileViewModel stuff may as well be separate, sorry about this!

Also, I noticed a major performance degradation with this, unnecessarily converting large binary files to UTF8 encoded strings to change for references to cubemap paths. Changed the UpdatePathReferences to optionally limit to a single extension, and passing in `.vmt` when called by the cubemap renaming stuff.

Issue #62, opened 2024-07-26
The output char buffer is too small to contain the decoded characters, encoding codepage '65001' 


dbampersand:
When loading a file with an entity lump that has characters with a codepoint greater than 127, the file will fail to parse correctly, outputting: "The output char buffer is too small to contain the decoded characters, encoding codepage '65001'"

This is because entities aren't actually compiled to UTF-8 but instead to 8-bit ASCII, for example if you put Ě (0xC49A) in an entity key it will compile to just E, but if you put a character that exists in extended ASCII like ÿ (0xFF) it won't get stripped. 

See these three TF2 jump maps for examples of this: https://filebin.net/rbh7fcpz4zmdiqo6 

For instance, on jump_4starters_b1_fix.bsp it crashes on the ç character (E7): https://i.imgur.com/QT5Xr2R.png

tsa96:
Huh, good spot. If this is being used everywhere in BSP (besides just ent lump) we perhaps should just make this encoding a static property of BspFile and use it everywhere. Will have a poke around in engine at some point to see if there's explicit mention of this.

PR #63, opened 2024-07-26
Entity parsing: UTF8 fix

dbampersand:
Fix for this issue: https://github.com/momentum-mod/lumper/issues/62

while (reader.BaseStream.Position < reader.BaseStream.Length && (x = (char)reader.ReadByte()) != '\0')

Fixes the crash and

stream.Write(Encoding.GetEncoding(28591).GetBytes($"{prop}\n"));

Fixes writing the file out again using the 'true' encoding - without this it writes corrupt bytes.

There might be a more elegant way to do the first fix, e.g:

public abstract class LumpReader(Stream input) : BinaryReader(input, encoding: Encoding.GetEncoding(28591), leaveOpen: true)

However I'm not 100% there aren't any UTF-8 strings in a BSP. I'll test more later.


tsa96:
I agree that setting the encoding property of the `LumpWriter` is a nicer approach, and according to OzxyBox a few months ago, UTF-8 is practically never used in Source 1. Using `28591` consistently throughout the codebase is likely a good shout, though I've relatively inexperienced with text encoding stuff. @BIRD311 do you have any thoughts on this?

Since I'm not merging this right now and it's your fork I'm going to left rebasing to you, I do  `git rebase -i upstream/main` then drop everything besides this commit.

tsa96 (approved):
Sorry this has taken so long, had other stuff on, merging now

PR #72, opened 2025-01-10
Improved CLI

tsa96:
Bunch of CLI improvements, plus some consolidation of build configs into a single file.
- Support for various Lumper.Lib features
  - Compression/decompression
  - Running workflows
  - Outputting to a different filename from original file will rename cubemaps
- Separates out parser logic from actual BSP-related code
- Pulls workflow loading into Lumper.Lib
- Uses NLog
- Better usage page
- Catches and prints Lumper.Lib errors with exit code handling
- Less arg checking, let CommandLine handle it

Usage output:
```
Lumper CLI v1.0.0.0 [https://github.com/momentum-mod/lumper]

ERROR(S):
  A required value not bound to option name is missing.

  Input BSP (pos. 0)    Required. Path to input BSP file
  -o, --output          Path to output BSP file. If not provided, and some action modifies the BSP,
                        the input BSP will be overwritten.
  -c, --compress        Save the output BSP file compressed.
  -C, --noCompress      Save the output BSP file uncompressed.
  -B, --skipBackup      (Default: false) Don't make a backup of the input BSP file before
                        overwriting it.
  -w, --workflow        Path to a JSON file containing a job workflow to apply to the BSP file.
  -j, --json            (Default: false) Output JSON summary of the BSP to <directory containing BSP
                        file>/<bsp file name>.json
  -J, --jsonPath        Output JSON summary of the BSP the given path, relative to current
                        directory. If given, --json can be omitted.
  -k, --jsonOpts        (Default: SortLumps, SortProperties, IgnoreOffset) Provide either flags (1 =
                        Sort Lumps, 2 = Sort Properties, 4 = Ignore Offsets), or comma-separator
                        list of names, e.g. sortproperties,ignoreoffsets.
  -v, ----verbose       (Default: false) Enable verbose output (include debug messages)
  --help                Display this help screen.
  --version             Display version information.

Without options, Lumper will simply read the BSP file and exit.
If any options provided cause modifications, the BSP will be output to the same path.
If you want to output to a different path, use the -o option.
If -c or -C are given, an output file will always be produced, even if unmodified.
If neither -c or -C are given and a BSP is output, BSP will be written in its current state.
```


src/Lumper.UI/ViewModels/Pages/Jobs/JobsViewModel.cs
```diff
@@ -147,37 +146,27 @@ public void RemoveSelectedJob()
...
 
-    public void Load(Stream stream)
+    public void LoadWorkflow(Stream stream)
     {
-        var serializer = new JsonSerializer();
-        using var sr = new StreamReader(stream);
-        using var reader = new JsonTextReader(sr);
-        List<Job>? jobs;
-        try
+        if (!Job.TryLoadWorkflow(stream, out List<Job>? workflow))
```
Panzerhandschuh, line 151: I think this can be `out var workflow`

src/Lumper.CLI/Program.cs
```diff
@@ -1,32 +1,182 @@
...
+
+        if (bspFile is null)
+            throw new InvalidDataException("Failed to load BSP file");
+
+        if (options.JobWorkflow is { } workflowPath)
+            RunWorkflow(bspFile, workflowPath);
+
+        if (_shouldWriteOutput || options.Compress || options.DontCompress)
+        {
+            DesiredCompression compression = DesiredCompression.Unchanged;
```
Panzerhandschuh, line 81: This should be `var compression = ...`

Panzerhandschuh:
`var` usage is inconsistent so I'm not gonna comment on everything. I'm pretty sure it's safe to use just about anywhere, unlike `auto` in C++.

src/Lumper.UI/ViewModels/Pages/Jobs/JobsViewModel.cs
tsa96: Me and BIRD discussed `var` rules a bit and our .editorconfig rules are
```editorconfig
csharp_style_var_for_built_in_types = false
csharp_style_var_when_type_is_apparent = true
csharp_style_var_elsewhere = false
```
[rules here](https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/style-rules/language-rules#c-style-rules)
Uunfortunately the rules are quite limited and it's not possible to get the exact behavior we'd like, this seemed to be the best available since 
- We'd rather under-utilise `var` then over-utilise - `var` can be hard to follow in some cases. We've had similar discussions with C++, I tend to *over*use `var`/`auto` and often results in harder to understand code
- It's nice to have a consistent ruleset, means this shouldn't need discussing in code review. I'd rather have a reliable rule that `dotnet format` automatically fixes than have to discuss this in code review.

src/Lumper.CLI/Program.cs
tsa96: As above. I'd personally prefer using `var` here, but dotnet code analysis rules don't allow you to make enum values use `var` without making hundreds of other variables use `var` that we don't want to.

tsa96:
> `var` usage is inconsistent so I'm not gonna comment on everything. I'm pretty sure it's safe to use just about anywhere, unlike `auto` in C++.

This isn't a matter of safety, it's a matter of readability - see above comment

tsa96:
PR here https://github.com/momentum-mod/lumper/pull/73

Panzerhandschuh (approved):

PR #85, opened 2025-01-28
Fix very small lumps not being compressed properly 

### LINKED ISSUE #74: Compressed maps sometimes detected as uncompressed
I'm not sure if this is a TS header parsing thing or a Lumper thing.

tsa96:
Closes #74

We were writing out lumps < 22 bytes large as uncompressed since the LZMA header actually causes the lump to get larger. That causes the BSP to fail our compression checks in the website repo.

`bspzip` doesn't seem to have our optimization, and it's such an insignificant amount of data anyway, no problem always writing the header.

### DIFF
--- a/src/Lumper.Lib/BSP/IO/LumpWriter.cs
+++ b/src/Lumper.Lib/BSP/IO/LumpWriter.cs
@@ -124,8 +124,7 @@ public abstract class LumpWriter(Stream output) : BinaryWriter(output, Encoding.
         long compressedLength = mem.Length;
 
         var writer = new BinaryWriter(mem);
-        const int lzmaId = ('A' << 24) | ('M' << 16) | ('Z' << 8) | 'L';
-        writer.Write(lzmaId);
+        writer.Write("LZMA"u8);
         writer.Write((int)uncompressedStream.Length);
         writer.Write((int)mem.Length - headerSize);
         writer.Write(lzmaStream.Properties);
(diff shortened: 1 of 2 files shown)


src/Lumper.Lib/BSP/IO/LumpWriter.cs
```diff
@@ -122,42 +121,20 @@ private LumpHeaderInfo WriteCompressed(Lump lump)
...
-            // bytes so useless for very small lumps.
-            if (uncompressedLength > 0)
-            {
-                Logger.Debug(
-                    "Compressed lump larger than uncompressed, skipping. "
-                        + $"(compressed: {mem.Length}, uncompressed: {uncompressedLength})"
-                );
-            }
+        var writer = new BinaryWriter(mem);
+        const int lzmaId = ('A' << 24) | ('M' << 16) | ('Z' << 8) | 'L';
```
Panzerhandschuh, line 127: A better way to write this:
`var lzmaId = Encoding.ASCII.GetBytes("LZMA");`

Panzerhandschuh (approved):

src/Lumper.Lib/BSP/IO/LumpWriter.cs
tsa96: Good shout, though just realized a UTF8 literal works here i.e. `var lzmaID = "LZMA"u8;`, which is a `ReadonlySpan<byte>` which a `BinaryWriter` will treat the actual same way as a `byte[]`.

Updating a few other usages of `Encoding.ASCII` to use `u8` literals.

Panzerhandschuh: I wasn't aware that was an option, neat.

PR #86, opened 2025-01-30
Show exception stack traces in Log Viewer

tsa96:
Small QoL thing, helpful when debugging.

### DIFF
--- a/src/Lumper.UI/Views/LogViewer/LogViewerView.axaml.cs
+++ b/src/Lumper.UI/Views/LogViewer/LogViewerView.axaml.cs
@@ -89,11 +89,16 @@ public partial class LogViewerView : ReactiveUserControl<LogViewerViewModel>
                 {
                     FontStyle = FontStyle.Italic,
                     FontWeight = FontWeight.Medium,
-                    Foreground = Brushes.Gray,
+                    Foreground = Brushes.LightGray,
                 }
             );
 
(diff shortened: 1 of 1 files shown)


src/Lumper.UI/Views/LogViewer/LogViewerView.axaml.cs
```diff
@@ -107,6 +112,14 @@ private void AddLogMessage(LogMessage logMessage)
                 );
 
                 Logs.Inlines.Add(new Run(ex.Message) { Foreground = Brushes.IndianRed });
+                if (ex.StackTrace is not null)
+                {
+                    Logs.Inlines.Add(new LineBreak());
+                    string indent = new(' ', 8 + 26);
```
Panzerhandschuh, line 118: The `8 + 26` seems arbitrary. Can you use `\t` instead?

Panzerhandschuh (approved):

src/Lumper.UI/Views/LogViewer/LogViewerView.axaml.cs
tsa96: I'm going for very specific padding here, not keen on tabs, and this makes more sense if you check the rest of the file. I could do a named constant for `26` but honestly with immediate-mode style UI like this I find the code easier to read when constants are written explicitly like this.

PR #87, opened 2025-01-30
Official Asset Detection and Removal

tsa96:


src/Lumper.UI/ViewModels/Shared/Pakfile/PakfileEntryViewModel.cs
```diff
@@ -34,6 +28,8 @@ protected MemoryStream DataStream
 
     public long? CompressedSize => BaseEntry.CompressedSize;
 
+    [ObservableAsProperty]
+    public string? Hash { get; }
     protected PakfileEntryViewModel(PakfileEntry baseEntry, BspNode parent)
```
tsa96: Tiny formatting blip due to commit ordering, isn't in final changes. Not worth fixing!

src/Lumper.UI/Views/Pages/VtfBrowser/VtfBrowserView.axaml
```diff
@@ -22,6 +23,38 @@
             <Button BorderBrush="#10FFFFFF" Background="#08FFFFFF" BorderThickness="1" Margin="4"
                     CornerRadius="4" Cursor="Hand" Command="{Binding OpenVtfImageWindow}" Padding="0">
               <Panel>
+                <StackPanel ZIndex="10" HorizontalAlignment="Right" VerticalAlignment="Top" Margin="8"
+                            Orientation="Horizontal" IsVisible="{Binding !!MatchingGameAssets}" Background="#88000000">
```
Panzerhandschuh, line 27: What does the `!!` do? I'm guessing it's checking if `MatchingGameAssets` is not null. I haven't seen this syntax before in C# and wasn't sure if it's necessary.

Panzerhandschuh (approved):

src/Lumper.UI/Views/Pages/VtfBrowser/VtfBrowserView.axaml
tsa96: Ah, I don't think I did this quite right. I wrote this a while back and went back over Avalonia converters yesterday, I was misunderstanding when I first wrote this, and didn't update.

Avalonia bindings aren't C# code (it's some very small DSL idk), where `!` can be used to negate a boolean, or type for which `Convert.ToBoolean` is defined. This *doesn't* apply for `null` though (very common to JS), and `MatchingGameAssets` is `List<PakfileEntryViewModel>?`, so we want to show it if the `List` is non-null and non-empty. I *think* doing `{Binding !!MatchGameAssets.Count}` will work correctly but I should double check. Or rather, I'm pretty sure my usage works okay (I've tested it ofc) but it's quite unclear what this is meant to be checking.

See here for Avalona data binding converter stuff https://docs.avaloniaui.net/docs/reference/built-in-data-binding-converters. Dunno if WPF has something similar. Also, for complex logic they're horrific, e.g. if you want to do an `and` or `or` converter: https://github.com/AvaloniaUI/Avalonia/blob/cc714a203c3a16310bd4104a38f8a525588f65a2/src/Avalonia.Themes.Default/TextBox.xaml#L25

Anyway, I'll update this first thing tomorrow.

PR #101, opened 2025-02-14
Fix errors parsing numbers on systems not using English

tsa96:
Fuckf loats

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -205,18 +205,6 @@ dotnet_diagnostic.IDE0059.severity = suggestion
 # Always provide the rule title and a justification for disabling.
 ##########################################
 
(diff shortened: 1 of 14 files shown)


Panzerhandschuh:
Why use `StringComparison.Ordinal` instead of `InvariantCulture` in some cases? I'm not very familiar with the differences.

Panzerhandschuh (approved):

tsa96:
> Why use `StringComparison.Ordinal` instead of `InvariantCulture` in some cases? I'm not very familiar with the differences.

There's some weird behaviour with `InvariantCulture` where some characters are considering equal even though their byte representation is different, plus `Ordinal` is much faster, and behaves the naive programmer-friendly way like `memcmp`/`strcmp` behave. Basically `Ordinal` says "I don't give a fuck about i18n whatsoever, just give me a byte-by-byte comparison.

PR #106, opened 2025-02-16
Implement loading BSPs by dragging over UI window

tsa96:
Closes https://github.com/momentum-mod/lumper/issues/18

Based off of https://github.com/momentum-mod/lumper/pull/19

### DIFF
--- a/src/Lumper.UI/Views/MainWindow.axaml.cs
+++ b/src/Lumper.UI/Views/MainWindow.axaml.cs
@@ -1,11 +1,14 @@
 namespace Lumper.UI.Views;
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reactive.Disposables;
 using Avalonia.Controls;
 using Avalonia.Controls.Primitives;
+using Avalonia.Input;
 using Avalonia.Interactivity;
+using Avalonia.Platform.Storage;
 using Avalonia.ReactiveUI;
 using Lumper.UI.Services;
 using Lumper.UI.ViewModels;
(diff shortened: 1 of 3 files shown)


src/Lumper.UI/Views/MainWindow.axaml.cs
```diff
@@ -77,4 +83,19 @@ private void PageButton_OnClick(object? _, RoutedEventArgs e)
...
+    }
+
+    private void Window_OnDragOver(object? _, DragEventArgs e)
+    {
+        bool hasBsps = FilterDraggedBspFiles(e).Count > 0;
+        e.DragEffects = hasBsps ? DragDropEffects.Link : DragDropEffects.None;
+    }
+
+    private List<IStorageFile> FilterDraggedBspFiles(DragEventArgs e) =>
+        e.Data.GetFiles()?.OfType<IStorageFile>().Where(file => file.Name.EndsWith(".bsp")).ToList() ?? [];
```
Panzerhandschuh, line 100: I think you need to add null propagation after `OfType<IStorageFile>()`:
`e.Data.GetFiles()?.OfType<IStorageFile>()?.Where(file => file.Name.EndsWith(".bsp")).ToList() ?? [];`

Panzerhandschuh (approved):

src/Lumper.UI/Views/MainWindow.axaml.cs
tsa96: In C#, unlike JS, `?.` short-circuits https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-

Also note that the nulluable reference type static analysis is very good at flagging this -- if you applied your suggestion the static analysis would raise a warning and release compiles would fail.

PR #110, opened 2025-02-22
Update Entity Rules

tsa96:
Updates from gsheet, fixes a bug where it getting the allow level wrong because I screwed up an enum in the python script

### DIFF
--- a/resources/EntityRules_Momentum.json
+++ b/resources/EntityRules_Momentum.json
@@ -15,14 +15,14 @@
         "Level": 3
     },
     "commentary_auto": {
-        "Level": 1,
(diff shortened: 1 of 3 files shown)


Panzerhandschuh:
Most of the denied entities have no impact on gameplay. Why do we have to hassle mappers to remove them?

tsa96:
> Most of the denied entities have no impact on gameplay. Why do we have to hassle mappers to remove them?
> 
> We should only deny entities that are actually causing problems.

Let's discuss this in #game-design, not here.

Panzerhandschuh (approved):

PR #145, opened 2025-04-20
Fix EntityIO viewmodels not updating after model changes

tsa96:
Fixing some of my gross entity MVVM garbage fire. This was causing UI displaying entity IO to fail to update after jobs or other Lumper.Lib based code modified them.

### DIFF
--- a/src/Lumper.UI/ViewModels/Shared/Entity/EntityLumpViewModel.cs
+++ b/src/Lumper.UI/ViewModels/Shared/Entity/EntityLumpViewModel.cs
@@ -171,24 +171,33 @@ public sealed class EntityLumpViewModel : BspNode, ILumpViewModel
             // Property updates on EL -> update on ELVM
             foreach (EntityPropertyViewModel propVm in entVm.Value.Properties)
             {
-                // Checking both references and underlying values. Jobs should generally
-                // just create an entirely new property (so ref compare is what matters here)
-                // so entityproperty ctor logic runs, but doesn't hurt to test for value
-                // changes as well.
-                switch (propVm)
+                // Checking both references and underlying values. Jobs can either create
+                // an entirely new property (so ref compare), or they can modify the existing
+                // property (so value compare)
+                if (
+                    propVm is EntityPropertyStringViewModel { EntityProperty: Entity.EntityProperty<string> strM } strVm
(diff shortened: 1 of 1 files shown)


src/Lumper.UI/ViewModels/Shared/Entity/EntityLumpViewModel.cs
```diff
@@ -171,24 +171,31 @@ public void UpdateViewModelFromModel()
...
-                // Checking both references and underlying values. Jobs should generally
-                // just create an entirely new property (so ref compare is what matters here)
-                // so entityproperty ctor logic runs, but doesn't hurt to test for value
-                // changes as well.
-                switch (propVm)
+                // Checking both references and underlying values. Jobs can either create
+                // an entirely new property (so ref compare), or they can modify the existing
+                // property (so value compare)
+                if (
+                    propVm is EntityPropertyStringViewModel { EntityProperty: Entity.EntityProperty<string> sM } sVm
```
Panzerhandschuh, line 178: These variables should probably use more descriptive names (`strVm` / `strViewModel`). `sVm` is a bit difficult to interpret.

Panzerhandschuh (approved):

src/Lumper.UI/ViewModels/Shared/Entity/EntityLumpViewModel.cs
tsa96: Fair enough. I think using m/vm shorthand in some cases makes sense, but I agree that it's quite hard to parse here.

tsa96:
Woops, this PR doesn't contain all the commits it should. On my phone atm, but I'll sort it out over next few days

src/Lumper.UI/ViewModels/Shared/Entity/EntityLumpViewModel.cs
tsa96: Doing `strVm`, don't want a million line breaks and codebase references viewmodels/models so much I think shorthand is fair.

tsa96:
Pretty sure I don't need this actually, incoming MVVM improvement stuff makes this logic a lot less gross. Leaving for now since I'm not totally sure

tsa96:
Yeah gonna do this differently

PR #147, opened 2025-04-23
Fix filenames getting trucated on export

### LINKED ISSUE #107: File and folder names being truncated on export
Image
Version 1 of [sj_santorini](https://staging.momentum-mod.org/maps/sj_santorini) can be used for testing

expectchaos:
If the common ancestor is the root path, we don't need to remove anything from the entry path. In that case commonParentPath.Length needs to be 0 but adding a slash makes it 1, which removes the first letter of the entry path. Therefore only add a slash if not in root directory.
Closes #107

### DIFF
--- a/src/Lumper.UI/ViewModels/Pages/PakfileExplorer/PakfileExplorerViewModel.cs
+++ b/src/Lumper.UI/ViewModels/Pages/PakfileExplorer/PakfileExplorerViewModel.cs
@@ -420,7 +420,10 @@ public sealed class PakfileExplorerViewModel : ViewModelWithView<PakfileExplorer
             return;
         }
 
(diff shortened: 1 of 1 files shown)


tsa96 (approved):
Yeah this makes sense, good work figuring this out!

tsa96:
Need to autoformat, should be `dotnet csharpier .`

tsa96:
Sorry I should've said, please avoid commits like "autoformat", just fixup the change into your original commit. You can do this with `git rebase -i HEAD~2`, then in the editor that appears, change the word "pick" on the second line to "fixup", then save and close the file -- that'll fixup the autoformat changes into the first commit. Then force push with `git push --force`.

tsa96 (approved):

PR #149, opened 2025-05-06
Add "VDC Reference" button to Entity Editor pages

fiveNinePlusR:
opens up the valve developer documentation for a given selected entity.

### DIFF
--- a/src/Lumper.UI/ViewModels/Pages/EntityEditor/EntityEditorTabViewModel.cs
+++ b/src/Lumper.UI/ViewModels/Pages/EntityEditor/EntityEditorTabViewModel.cs
@@ -12,34 +12,36 @@ using Lumper.Lib.Bsp.Struct;
 using Lumper.UI.Services;
 using Lumper.UI.ViewModels.Shared.Entity;
 using ReactiveUI;
+using ReactiveUI.Fody.Helpers;
 
-public abstract class EntityEditorTabViewModel
+public abstract class EntityEditorTabViewModel : ViewModel
 {
-    public abstract EntityViewModel Entity { get; }
+    public EntityViewModel Entity { get; protected init; } = null!;
 
-    public abstract string Name { get; protected set; }
+    [Reactive]
+    public string Name { get; protected set; } = null!;
 
     public bool IsPinned { get; set; }
 
     public abstract bool Pinnable { get; }
 }
 
-public class EntityEditorTabSingleEntityViewModel(EntityViewModel entity) : EntityEditorTabViewModel
+public class EntityEditorTabSingleEntityViewModel : EntityEditorTabViewModel
 {
-    public override EntityViewModel Entity { get; } = entity;
-
-    public override string Name { get; protected set; } = entity.Classname;
-
     public override bool Pinnable => true;
+
+    // public override string DocumentationUri => "https://developer.valvesoftware.com/wiki/" + Entity.Classname;
+    public EntityEditorTabSingleEntityViewModel(EntityViewModel entity)
+    {
+        Entity = entity;
+
+        entity.WhenAnyValue(x => x.Classname).Do(x => Name = x).Subscribe();
+    }
 }
 
 public sealed class EntityEditorTabMultipleEntityViewModel : EntityEditorTabViewModel, IDisposable
 {
-    public override EntityViewModel Entity { get; }
-
     public List<EntityViewModel> RealEntities { get; }
-
-    public override string Name { get; protected set; }
 
     public override bool Pinnable => false;
 
--- a/src/Lumper.UI/ViewModels/Pages/EntityEditor/EntityEditorTabViewModel.cs
+++ b/src/Lumper.UI/ViewModels/Pages/EntityEditor/EntityEditorTabViewModel.cs
@@ -19,7 +19,10 @@ public abstract class EntityEditorTabViewModel : ViewModel
     public EntityViewModel Entity { get; protected init; } = null!;
 
     [Reactive]
-    public string Name { get; protected set; } = null!;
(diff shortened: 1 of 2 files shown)


tsa96:
Can you post a screenshot please?

src/Lumper.UI/Views/Pages/EntityEditor/EntityEditorView.axaml
```diff
@@ -244,10 +244,14 @@
...
-                <Button Command="{Binding Entity.AddString}">Add KeyValue</Button>
-                <Button Command="{Binding Entity.AddIo}">Add IO</Button>
-              </StackPanel>
+              <DockPanel>
+                <Button Command="{Binding Entity.AddString}" HorizontalAlignment="Left" Margin="4 0 ">Add KeyValue</Button>
+                <Button Command="{Binding Entity.AddIo}" HorizontalAlignment="Left" Margin="4 0">Add IO</Button>
+                <HyperlinkButton NavigateUri="{Binding DocumentationUri}"
+                                 Content="VDC Reference"
+                                 HorizontalAlignment="Right"
+                                 IsVisible="{Binding DocumentationAvailable}"/>
```
tsa96, line 253: I think you can avoid having to define `DocumentationAvailable` by using `ObjectConverters.IsNotNull` https://docs.avaloniaui.net/docs/reference/built-in-data-binding-converter, just make DocumentationUri a `string?` then return null for the multi-tab VM.

src/Lumper.UI/ViewModels/Pages/EntityEditor/EntityEditorTabViewModel.cs
```diff
@@ -31,6 +36,11 @@ public class EntityEditorTabSingleEntityViewModel(EntityViewModel entity) : Enti
     public override string Name { get; protected set; } = entity.Classname;
 
     public override bool Pinnable => true;
+
+    public override string DocumentationUri => "https://developer.valvesoftware.com/wiki/" + Entity.Classname;
```
tsa96, line 40: Classname can be the `MissingClassname` string, in which case should be null.
```c#
public override string DocumentationUri =>
  Entity.Classname == Entity.MissingClassname ? $"https://developer.valvesoftware.com/wiki/{Entity.Classname}" : null;
```

tsa96 (changes requested):
When you change the classname to a different value, does the URL change? I'm not sure when HyperlinkButton will evaluate the `DocumentationUri` expression.

tsa96:
Need to autoformat, `dotnet csharpier .`. Please amend that and everything addressed in review into your original commit!

fiveNinePlusR:
image

tsa96:
Not when selected, I mean if you change the classname property of that entity whilst you're editing it. `DocumentationUrl` is not a reactive property, so Avalonia might not update the value when it changes.

fiveNinePlusR:
ah yeah, that does not update the classname dynamically

tsa96:
Okay one sec, lemme write that.

fiveNinePlusR:
i did an autoformat with `dotnet csharpier .` and it changed 119 files so i'm not sure about that

tsa96:
This is untested, if there's problems and you can't figure it out I'll sort it out tomorrow.

```c#
public abstract class EntityEditorTabViewModel : ViewModel
{
    public EntityViewModel Entity { get; protected init; } = null!;

    public abstract string Name { get; set; }

    public bool IsPinned { get; set; }

    public abstract bool Pinnable { get; }

    public abstract string? DocumentationUri { get; }
}

public class EntityEditorTabSingleEntityViewModel : EntityEditorTabViewModel
{
    [ObservableAsProperty]
    public override string Name { get; set; } = null!;

    public override bool Pinnable => true;

    [ObservableAsProperty]
    public override string? DocumentationUri { get; }

    public EntityEditorTabSingleEntityViewModel(EntityViewModel entity)
    {
        Entity = entity;

        entity.WhenAnyValue(x => x.Classname).ToPropertyEx(this, x => x.Name);
        entity
            .WhenAnyValue(x => x.Classname)
            .Select(x => x != EntityViewModel.MissingClassname ? $"https://developer.valvesoftware.com/wiki/{x}" : null)
            .ToPropertyEx(this, x => x.DocumentationUri);
    }
}

public sealed class EntityEditorTabMultipleEntityViewModel : EntityEditorTabViewModel, IDisposable
{
    public List<EntityViewModel> RealEntities { get; }

    public override string Name { get; set; }

    public override bool Pinnable => false;

    public override string? DocumentationUri => null;

... rest of file is the unchanged
```
this should also makes it so changing Classname will update the name of the tab.

tsa96:
> i did an autoformat with `dotnet csharpier .` and it changed 119 files so i'm not sure about that

For me I get a "Formatted 124 files in ...ms", that hasn't actually modified them. Check in Git whether they've actually changed! If someone's gone wrong, make sure you're running in the root directory of the project, and might need to run `dotnet tool restore`.

fiveNinePlusR:
i'm not sure what's up but 

```
PS C:\Users\user\Documents\projects\lumper_project\lumper> dotnet csharpier .
Formatted 119 files in 244ms.
PS C:\Users\user\Documents\projects\lumper_project\lumper> dotnet tool restore
Tool 'csharpier' (version '0.30.1') was restored. Available commands: dotnet-csharpier

Restore was successful.
PS C:\Users\user\Documents\projects\lumper_project\lumper> dotnet csharpier .
Formatted 119 files in 1269ms.
```

in the middle there i reset the changes with git and it redid the 119 files

tsa96 (approved):
Hijacking this as discussed in Discord. I need to stop doing this with people's PRs, Avalonia is just really hard to get right!

Really appreciate the work though, keep you as author on the main commit, though I modified it a little.

PR #152, opened 2025-05-11
Gamelump corruption fixes

### LINKED ISSUE #151: dprp gamelump adding junk data on read/write
https://discord.com/channels/235111289435717633/998769771451519088/1370064960553615472

expectchaos:
Fixes #151
Plus another small issue I noticed, when the sprp has no entries it should still contain the 3 entry counts (which are all 0), this however did not get written on save since the lump is considered empty. Didn't seem to break anything but the data of the lump probably shouldn't arbitrarily be deleted like that.

### DIFF
--- a/src/Lumper.Lib/BSP/Lumps/GameLumps/Sprp.cs
+++ b/src/Lumper.Lib/BSP/Lumps/GameLumps/Sprp.cs
@@ -80,8 +80,5 @@ public class Sprp(BspFile parent) : ManagedLump<GameLumpType>(parent)
         }
     }
 
-    public override bool Empty =>
-        (StaticProps is null || StaticProps.Empty)
-        && (StaticPropsDict is null || StaticPropsDict.Empty)
-        && (StaticPropsLeaf is null || StaticPropsLeaf.Empty);
+    public override bool Empty => StaticProps is null && StaticPropsDict is null && StaticPropsLeaf is null;
 }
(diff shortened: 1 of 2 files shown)


src/Lumper.Lib/BSP/Lumps/GameLumps/Sprp.cs
```diff
@@ -80,8 +80,5 @@ public override void Write(Stream stream, IoHandler? handler = null, DesiredComp
         }
     }
 
-    public override bool Empty =>
-        (StaticProps is null || StaticProps.Empty)
-        && (StaticPropsDict is null || StaticPropsDict.Empty)
-        && (StaticPropsLeaf is null || StaticPropsLeaf.Empty);
+    public override bool Empty => StaticProps is null && StaticPropsDict is null && StaticPropsLeaf is null;
```
tsa96, line 83: Yeah, I can see why this is a problem, but if so, am I right in thinking that StaticProps, StaticPropsDict ad StaticPropsLeaf are never null? That seems to make false, but in that case we should update the types to be non-nullable, and remove the comment on line 11. Then this method would become simply `=> false`.

I have to admit I still barely understand game lump stuff (BIRD did the original implementation of this), might be misunderstanding. If I'm right about them being non-null then let's leave a comment on this method explaining why it's always false.

tsa96 (changes requested):
Nice work figuring this out!

src/Lumper.Lib/BSP/Lumps/GameLumps/Sprp.cs
expectchaos: They can be null if the sprp header has length 0. That shouldn't really happen after fixing this but I feel it's safer to keep it. Considering that the map still loaded fine in game, it might be possible that there is a map like that out there.

tsa96 (approved):

Issue #177, opened 2025-06-04
Asset removal results in static props with error textures


tsa96:
Asset removal on surf_mesa results in error textures in the ceiling. Need to remove from static prop lump.

PR #178, opened 2025-06-04
Remove deleted static props during RemoveAssetJob

tsa96:
Closes https://github.com/momentum-mod/lumper/issues/177

Okay this is a silly PR, I just realised that we probably shouldn't remove these since they're important to gameplay and we'll do the outlined collision mesh-based approach before 1.0.0. Keeping the code in case every useful, but disabled unless something enables it on the job's properties.

### DIFF
--- a/src/Lumper.UI/Views/Pages/Jobs/RemoveAssetJobView.axaml
+++ b/src/Lumper.UI/Views/Pages/Jobs/RemoveAssetJobView.axaml
@@ -27,6 +27,12 @@
             </DataTemplate>
           </ItemsControl.ItemTemplate>
         </ItemsControl>
+        <StackPanel Orientation="Horizontal" Spacing="8">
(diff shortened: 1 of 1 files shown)


src/Lumper.Lib/Jobs/RemoveAssetJob.cs
```diff
@@ -102,4 +107,22 @@ public override bool Run(BspFile bsp)
...
         }
     }
+
+    private static void RemoveStaticProp(BspFile bsp, string path)
+    {
+        Sprp? sprp = bsp.GetLump<GameLump>().GetLump<Sprp>();
+        if (sprp is null)
+            return;
+
+        int idx =
```
Panzerhandschuh, line 126: I keep forgetting how this works, but does the `StaticPropsDict` null conditional short circuit the call to `FindIndex`?

PR #210, opened 2025-10-15
Add SURF_SKYOCCLUSION surface flag to SURF_SKY(2D) job

SCell555:
Allows old maps to use our skybox occlusion we have on tools/toolsky(2d) materials without recompiling the map.
 
Also fixes definition of Surface flags enum not being actual flags.
 
(Also no clue what I'm doing with avalonia lmao)
(Didn't have time to test this yet on actual bsp, since I was writing this at 1am lol)


SCell555:
Just tested this with real BSP. Found a spot where I could see geometry through skybox. Run it through lumper with this job and saved. And checked the same spot and I could only see sky. And it seems to have worked properly.

tsa96:
Legend

Issue #124, opened 2025-03-07
Lumper does not detect Momentum Mod's original textures


Natanxp2:
Various textures that come with the game are not detected by lumper and the "Remove Game Assets" job does not remove them. This can lead to mappers accidentally packing them. Since those textures are downloaded with the game itself they should not be packed in any map.

Example of textures that should not be packed ( checkpoint, move limit, no grenade, etc... )
![Image](https://github.com/user-attachments/assets/60cf1947-d245-4473-8485-5c03d55692e5)

tsa96:
This will take a lot of work and need stuff in Momentum's game CI, low priority and really doesn't matter if some maps end up including our own stuff. Will be higher priority when we get PolyHaven stuff in.

PR #222, opened 2025-12-10
Various asset removal fixes

### LINKED ISSUE #198: Cubemaps getting deleted on asset removal
Some maps have not built cubemaps, so they use the default cubemap that is generated on compile. That cubemap gets mistakenly detected as a portal2 asset "materials/cubemaps/cubemap_glass_01.vtf" and deleted on asset removal. This breaks the cubemaps on models since they don't use vmt's but look for the cubemap file directly. And path refactoring forces the portal cubemap to be used in vmt's so the hdr cubemaps no longer work.

Example map: [surf_not_so_neb](https://gamebanana.com/mods/332411)
In this case the hdr cubemap is shiny and reflective while the non hdr cubemap is the default generated one and non reflective. The generated cubemap gets deleted and vmt's are refactored. The portal cubemap is used rather than the hdr one and ramps are no longer shiny.

expectchaos:
Added hl2e2 and portal2 to asset origin priority list. Since ep2 is shipped with the game now it should take priority over other mounted games. While portal2 assets aren't desirable, they should still be preferred to games that aren't mounted, like portal1. There are some assets shared between portal1 and 2 but at a different path and portal1 would always get chosen due to appearing first in asset manifest.

Prevented the deletion of cubemaps by asset removal. Should never be removed. Fixes #198.

Made it so asset removal only has mounted games checked by default. There have been too many instances of people running asset removal with default settings and having to repack deleted assets after cause they were from non-mounted games like l4d2 or dod. If any remaining assets are problematic they can be caught in content approval and addressed individually.

### DIFF
--- a/src/Lumper.Lib/AssetManifest/AssetManifest.cs
+++ b/src/Lumper.Lib/AssetManifest/AssetManifest.cs
@@ -51,13 +51,10 @@ public static class AssetManifest
     public static HashSet<string> Origins { get; } = [];
 
     /// <summary>
-    /// Origins we prioritise when renaming assets that match a hash but not the original filename,
-    /// or there's multiple officials with the same hash, where one path matches but a more
-    /// accessible game has the same asset with a different path.
-    /// In both case, we update references to the most accessible path everywhere to avoid missing
-    /// textures as best we can. This is biased towards Momentum!
+    /// List of games that are mounted by default. The order determines game priority in asset removal and required games.
+    /// This is biased towards Momentum!
     /// </summary>
-    public static readonly List<string> RenamedOriginPriority = ["hl2", "tf2", "css", "csgo"];
+    public static readonly List<string> MomentumMountedOrigins = ["hl2", "tf2", "css", "csgo"];
 
     /// <summary>
     /// Preload the asset manifest. UI wants this, CLI doesn't. Don't call from the UI thread!
--- a/src/Lumper.Lib/AssetManifest/AssetManifest.cs
+++ b/src/Lumper.Lib/AssetManifest/AssetManifest.cs
@@ -54,7 +54,16 @@ public static class AssetManifest
     /// List of games that are mounted by default. The order determines game priority in asset removal and required games.
     /// This is biased towards Momentum!
     /// </summary>
-    public static readonly List<string> MomentumMountedOrigins = ["hl2", "tf2", "css", "csgo"];
+    public static readonly List<string> MomentumMountedOrigins =
+    [
+        "hl2",
+        "hl2e2",
+        "hl2e1",
+        "tf2",
+        "css",
+        "csgo",
+        "portal2",
+    ];
 
     /// <summary>
     /// Preload the asset manifest. UI wants this, CLI doesn't. Don't call from the UI thread!
--- a/src/Lumper.Lib/Jobs/RemoveAssetJob.cs
+++ b/src/Lumper.Lib/Jobs/RemoveAssetJob.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Lumper.Lib.AssetManifest;
 using Lumper.Lib.Bsp;
 using Lumper.Lib.Bsp.Lumps.BspLumps;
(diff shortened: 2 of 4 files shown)


expectchaos:
> Currently packed cubemaps show up as official assets in the Map Summary, so that would need to be changed as well

That I believe would require to update the asset manifest and remove the texture it is falsely flagging against. I don't think that is something I am able to do. When Tom was removing assets from the manifest a while back, I did ask to remove this too but it's still flagging it. Some assets like cs_havana/white and cs_italy/black are also getting flagged despite supposedly being removed so idk.

AshenCinders:
@tsa96 what to do here?

tsa96:
> > Currently packed cubemaps show up as official assets in the Map Summary, so that would need to be changed as well
> 
> That I believe would require to update the asset manifest and remove the texture it is falsely flagging against. I don't think that is something I am able to do. When Tom was removing assets from the manifest a while back, I did ask to remove this too but it's still flagging it. Some assets like cs_havana/white and cs_italy/black are also getting flagged despite supposedly being removed so idk.

Shall we just hardcode a list of paths to ignore in that case, and ignore anything that matches?

tsa96 (changes requested):

expectchaos:
> Shall we just hardcode a list of paths to ignore in that case, and ignore anything that matches?
That list is gonna grow really large as more and more assets get shipped with the game. Would need an efficient way to store and read all of them, that is kinda what the manifest is for. The current approach of removing shipped assets from the manifest isn't ideal either cause it also stops them from being removed by asset removal. Since they are shipped they don't need to be packed. Adding them to the manifest as "mom" assets for example would allow them to be removed but ignored in required games like hl2 is currently. This also intersects with #124 and I think is out of scope of this PR, I just wanted to fix cubemaps getting deleted which can cause various issues and is very hard to detect and fix after the fact.

For the cubemap texture we do want to both not remove it and not show up in required games, and to me just removing it from the manifest seems the simplest, but then again I don't know how the manifest is generated and how hard that would be. Would do it myself if I knew how.

tsa96:
Current skip setup looks like this
`generate-manifest.sh`
```bash
#!/bin/bash
# Uses `verifier` to generate an RSV file for Lumper. Writes out to /bin/win64/verifier_index.rsv for some reason.
# Just rename to assetmanifest.rsv and include with Lumper
#
# https://github.com/StrataSource/verifier
#
# usage: verifier.exe [--new-index] [--root [root]] [--skip-archives] [--exclude
# excluded [excluded ...]] [--include included [included ...]] [--index
# [index-loc]] [--overwrite] [--ui-report] [--help]
#
# A tool used to verify a game's install. v0.3.1
#
# optional arguments:
#   --new-index                 Creates a new index file.
#   --root [root]               The engine root directory.
#   --skip-archives             Don't parse files stored in VPKs, parse the entire
#                               VPK instead.
#   -e, --exclude excluded [excluded ...]
#                               RegExp pattern(s) to exclude files when creating
#                               the index.
#   --include included [included ...]
#                               RegExp pattern(s) to include files when creating
#                               the index. If not present, all files not matching
#                               an exclusion will be included.
#   -i, --index [index-loc]     The index file to use.
#   --overwrite                 Do not ask for confirmation for overwriting an
#                               existing index.
#   --ui-report                 Use UI report mode logging.
#   -h, --help                  Display this help message and exit.
exclude=".*$(cat ignored_subdirs.txt | tr '\n' '%' | sed -e 's/%/\/.* .*/g')/.*"
echo $exclude
./verifier.exe --new-index --include ".*materials/.+" ".*models/.+" ".*particles/.+" --exclude $exclude
 
 # TODO: IGNORE HLSRC SUBDIR whatever
 ```
 `ignored_subdirs.txt`
 ```
 paint
backpack
player/items
weapons/customization
stickers
inventory_items
workshop
materials/gm_forest
materials/tools
materials/cs_italy/white
materials/cs_italy/black
materials/cs_havana/white
materials/effects/black
materials/lights/white
materials/details/dt_smooth01
materials/cubemaps/cubemap_glass_01
```

tsa96:
I could get you access to this specific directory on our Nextcloud instance if you like, which contains all the VPKs in question. Sorry I'm being sloppy with that, don't have much time atm.

tsa96 (approved):

tsa96:
Okay looks good, sorry this has taken so long!

PR #226, opened 2026-05-07
fix(ui): don't create file / directory if user cancelled

### LINKED ISSUE #224: Clicking cancel on creating file/directory still creates it

kusabana:
fixes #224


tsa96 (approved):
Thanks!

PR #229, opened 2026-05-07
fix(ui): seperate entity lump save and editor teardown

### LINKED ISSUE #202: Raw Entities tab doesn't update lump after first save
### Describe the bug

The Raw Entities tab correctly updates the lump when saving the first time, but on further changes re-saving no longer updates the lump. Switching to a different tab may fix the issue.

### How To Reproduce

1. Open a map in Lumper and switch to the Raw Entities tab
2. Make any change to the data and save
3. Notice the log shows "Updated entity lump" and the BSP has the new data
4. Make another change to the data and save
5. Notice the log doesn't show "Updated entity lump" and the BSP has the previously saved data

kusabana:
fixes #202


tsa96 (approved):
Nicely done, thank you!

PR #232, opened 2026-05-17
fix(lib): avoid replacing %OUTPUT% when writing to input file

kusabana:


tsa96 (approved):
Apologies for taking so long to get to this, will review all your stuff either today or tomorrow!

PR #236, opened 2026-05-23
fix(ui): update pakfile tree size when entry data changes

### LINKED ISSUE #234: File size in Pakfile Explorer doesn't update
If i apply an update such as resize or re-encode, the size will update in the inspector, but not in the file list.

kusabana:
I'm not confident this is the best approach as i don't have any experience with DynamicData, so i would appreciate any input.

Fixes #234

### DIFF
--- a/src/Lumper.UI/ViewModels/Pages/PakfileExplorer/PakfileTreeViewModel.cs
+++ b/src/Lumper.UI/ViewModels/Pages/PakfileExplorer/PakfileTreeViewModel.cs
@@ -44,7 +44,7 @@ public class PakfileTreeViewModel
                     }
                 }
 
-                Root.RecalculateSize();
+                Root.RecalculateSizes();
             });
     }
 
@@ -145,7 +145,6 @@ public class PakfileTreeNodeViewModel : ViewModel
             if (existing is not null)
             {
                 existing.AddInternal(value, path[1..]);
-                existing.Size += size;
                 return;
             }
 
@@ -196,7 +195,6 @@ public class PakfileTreeNodeViewModel : ViewModel
         {
             node.RemoveRecursive(path[1..]);
 
-            node.RecalculateSize();
             // Delete directory is no children is empty.
             // Even though zips don't have actual directories we *do* let you add
             // empty directories (so you can add stuff to them), so maybe this isn't quite the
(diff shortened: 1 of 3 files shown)


src/Lumper.UI/ViewModels/Pages/PakfileExplorer/PakfileTreeViewModel.cs
```diff
@@ -101,7 +99,21 @@ public class PakfileTreeNodeViewModel : ViewModel
...
     {
-        Size = Leaf?.UncompressedSize ?? 0;
+        Children = [];
+
+        Children
+            .ToObservableChangeSet()
+            .AutoRefresh(c => c.Size)
+            .ToCollection()
+            .Select(coll => (long?)coll.Sum(c => c.Size ?? 0))
+            .ToPropertyEx(this, x => x.Size);
```
tsa96: I'm afraid I'm seeing some quite major performance issues. Testing with a BSP with ~1000 items in the pakfile (mostly just .vhv in top level) the `PakfileExplorerViewModel` constructor takes ~2500ms, whereas on the prior commit the same file takes ~400ms. I haven't profiled but ultimately ChangeSet creation is pretty expensive.

A little on the fence about this, on one hand it's aaaaalmost an acceptable perf hit, on the other it's a shame to do this much compute given this changes relatively infrequently.

I haven't tried it, but wouldn't a simpler approach here be to recurse up a file's parent nodes in `RecalculateSize`?

tsa96 (changes requested):

src/Lumper.UI/ViewModels/Pages/PakfileExplorer/PakfileTreeViewModel.cs
tsa96: Oh and sorry again for the slow review!

kusabana: > A little on the fence about this, on one hand it's aaaaalmost an acceptable perf hit, on the other it's a shame to do this much compute given this changes relatively infrequently.

It's definitely not acceptable, I hadn't considered the performance implications and got a bit lost in DynamicData stuff (and didn't test on large pakfiles)

> I haven't tried it, but wouldn't a simpler approach here be to recurse up a file's parent nodes in `RecalculateSize`?

I tested this, and had issues with directory drag and drop. I've pushed a similar approach (recalculating entire tree on changes) and the performance impact seems negligible to me (~1ms with 5000 entries).

With this alternate approach, there's not much justification for the rest of the `PakfileTreeNodeViewModel` changes as part of this PR other than it just "being nicer" (in my opinion)
Let me know what you think.

tsa96: Great, I'll check that out tomorrow. I agree that having `Children` by non-null all the time is sometimes more ergonomic, though I think having them nullable is a nice way to model the fact a node is a genuine entry in the zip iff `Children` is null. On balance, I'd say best to keep this change small and leave the other `PakfileTreeNodeViewModel` changes.

kusabana: > On balance, I'd say best to keep this change small and leave the other `PakfileTreeNodeViewModel` changes.

I agree, I've removed the `PakfileTreeNodeViewModel` changes.

tsa96 (approved):
Great stuff, thank you for this!

PR #238, opened 2026-07-07
Add dropdowns to entity editor

Natanxp2:
# Adds FGD Compiler
Using roslyn source generator pre-compiles momentum.fgd to a readonly dictionary in c#

# Adds dropdowns
Every relevant property in entity editor has associated suggestions that are dynamically linked

# Notes
I tried to research TemplatedControls vs UserControls and it seems like TemplatedControls allow for more customization and are generally preferred for complicated componets. I could probably convert it to UserControl if you prefer that here

Getting targetnames for IO target or "target" property has been the most complicated part of this. I'm sure I did something wrong there and possibly used a completely incorrect approach.

Selecting the classname through the dropdown with arrow keys is janky but I'm not sure what to do with that, using the mouse or typing it in works perfectly fine


Panzerhandschuh:
I reviewed this with Opus 4.8. Let me know if this is helpful. I could also have Opus push fixes for these issues if you'd like.

Review of #238 against head `c9f768e`. Ten correctness findings; the first three cause silent data loss, so I'd treat those as blocking. Everything else is a real bug but not a save-corruption risk.

I built both `Lumper.Lib` and `Lumper.UI` — the FGD source generator runs clean on the current `momentum.fgd`, so nothing here is a build break today.

---

## Blocking — silent data loss

### 1. `IOKey` never writes through to the model, so I/O key renames are discarded on save

[`EntityPropertyViewModel.cs:260-268`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/ViewModels/Shared/Entity/EntityPropertyViewModel.cs#L260-L268)

The XAML moved from `<TextBox Text="{Binding Key}"/>` to `Text="{Binding IOKey, Mode=TwoWay}"`, but the new setter only touches a backing field:

```csharp
public string IOKey
{
    get => _ioKey;
    set
    {
        _ioKey = value;      // no `Property.Key = value;`
        OnValueChanged();    // calls MarkAsModified() — file is now dirty
    }
}
```

The base `Key` setter does `_key = value; Property.Key = value;`. `IOKey` does neither. Rename an output `OnPressed` → `OnDamaged`, save, reload: it's `OnPressed` again. The BSP was marked dirty and saved, just without the change.

Three knock-on effects:

- No `this.RaisePropertyChanged()`, so `PullChangesFromModel` can never refresh the box.
- `_key` and `_ioKey` diverge, so `MatchKey` filtering, `MemberwiseEquals` (`o.Key == Key`, line 399), and `DeleteProperty`'s `propVm.Key == "targetname"` check all read the stale key.
- `EntityEditorTabViewModel` matches multi-select edits by `.Key` ([lines 118](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/ViewModels/Pages/EntityEditor/EntityEditorTabViewModel.cs#L118) / [164](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/ViewModels/Pages/EntityEditor/EntityEditorTabViewModel.cs#L164)) and writes `realProp.Key` ([line 185](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/ViewModels/Pages/EntityEditor/EntityEditorTabViewModel.cs#L185)), so multi-entity edits target the wrong property.

The cleanest fix is to delete `_ioKey`/`IOKey` entirely and bind the XAML back to the inherited `Key`, which already does the model write, the notification, and the classname/targetname bookkeeping. If the reason for the shadow property was to dodge the base setter's `classname`/`targetname` side effects, the right move is to lift those side effects out of the setter rather than route around it.

### 2. `CalculateBitfieldSum` overwrites `spawnflags`, destroying bits it doesn't know about

[`ExtendedAutoCompleteBox.axaml.cs:300-305`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/Controls/ExtendedAutoCompleteBox.axaml.cs#L300-L305)

```csharp
if (long.TryParse(Text, out long currentValue) && currentValue == sum)
    return;

_isUpdatingText = true;
Text = sum.ToString(CultureInfo.InvariantCulture);   // unconditional clobber
_isUpdatingText = false;
```

`sum` is built only from items that are currently checked. Anything the checkbox state can't represent is silently dropped. Two reachable paths:

**Unknown bits.** An entity has `spawnflags = 1057`, but the FGD's choices for that classname only declare bits `1` and `32`. Entering bitfield mode ticks the known bits, `sum` comes out `33`, `33 != 1057`, and `Text = "33"` flows through the TwoWay binding into `_property.Value`. Bit `1024` is gone *on load*, before the user touches anything. Real FGDs routinely omit engine-level flags, so this isn't hypothetical.

**Suggestions arriving after the mode flip.** The ctor subscription sets `IsBitfield` first, *then* raises `ValueSuggestions` — which is a computed getter allocating fresh items with `IsChecked == false`. So retyping a key to `spawnflags` flips `IsBitfieldMode` against the *old* suggestion list; the new list then arrives via `OnSuggestionsChanged`, which never re-runs `InitializeBitfieldCheckboxes`. Every box renders unchecked while `Text == "3"`. Tick flag `8` and you get `Text = "8"` — flags `1|2` destroyed.

Also worth noting: the accumulator uses `sum += val` rather than `sum |= val`, so any duplicated bit value double-counts.

Suggested fix: preserve unknown bits explicitly. Capture the original value's unrepresented bits on entering bitfield mode and OR them back in, and don't write `Text` at all unless a checkbox actually changed (as opposed to the mode being initialized).

### 3. `_isInitializing` is never reset on the early-return path

[`ExtendedAutoCompleteBox.axaml.cs:252-257`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/Controls/ExtendedAutoCompleteBox.axaml.cs#L252-L257)

```csharp
private void InitializeBitfieldCheckboxes()
{
    _isInitializing = true;

    if (Suggestions == null || string.IsNullOrEmpty(Text))
        return;                                    // flag stays true, forever
```

Reached whenever the bound property has an empty value (an entity with `spawnflags = ""`), or if Avalonia applies the `IsBitfieldMode` binding before the `Suggestions` binding. From then on the flag is stuck true for the control's lifetime, and every checkbox toggle hits `if (_isInitializing) return;` at [line 282](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/Controls/ExtendedAutoCompleteBox.axaml.cs#L282). The checkboxes visibly toggle and `Text` is never written — the flags editor silently does nothing. This will land as an intermittent "flags don't save" report.

`try { … } finally { _isInitializing = false; }`, or drop the flag and unsubscribe/resubscribe `OnItemPropertyChanged` around the mutation loop.

---

## Correctness

### 4. Throttled projection runs on the taskpool and enumerates a UI-mutated collection

[`TargetnameIndexViewModel.cs:41-51`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/ViewModels/Shared/Entity/TargetnameIndexViewModel.cs#L41-L51)

`Throttle(200ms, RxApp.TaskpoolScheduler)` means the downstream `.Select(...)` executes on a threadpool thread — `ObserveOn(RxApp.MainThreadScheduler)` is only applied *after* it. That `Select` reads `e.Targetname` and `e.Classname`, both of which route through [`EntityViewModel.FindProperty`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/ViewModels/Shared/Entity/EntityViewModel.cs#L104-L107), which enumerates the live `Properties` collection.

Edit a targetname, then delete any property within the 200 ms window: `DeleteProperty` does `Properties.Remove(...)` on the UI thread while the taskpool is mid-enumeration. `InvalidOperationException: Collection was modified`, thrown inside an Rx subscription with no error handler. The same window also yields torn reads pairing a new targetname with a stale classname.

Move `ObserveOn(MainThreadScheduler)` above the `Select`, or throttle on the main-thread scheduler.

### 5. `PullChangesFromModel` bypasses `RaiseTargetnameChanged()`

[`EntityLumpViewModel.cs:174`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/ViewModels/Shared/Entity/EntityLumpViewModel.cs#L174) (also 217-218, 234-235, 243)

`Targetname` is computed as `FindProperty("targetname")`, and the only notifications come from `AddProperty`, `DeleteProperty`, and the `Key`/`Value` setters. But `PullChangesFromModel` mutates `entVm.Properties` directly, bypassing all four. After a Stripper or entity job adds or deletes a `targetname`, `Targetname`/`ClassAndTargetname` never re-raise — so the entity tree shows a stale name, *and* `TargetnameIndexViewModel`'s `.AutoRefresh(e => e.Targetname)` never fires. Every `target` and `TargetEntityName` autocomplete then offers names that no longer exist and omits the new ones.

This is the structural cost of the new manual-invalidation contract: the very first pre-existing caller already violates it, and any future mutation path (undo/redo, paste, bulk edit) will forget by default. Deriving `Targetname` reactively from `Properties` would make it unforgettable.

### 6. Event handlers on a singleton collection are never unsubscribed

[`ExtendedAutoCompleteBox.axaml.cs:157-199`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/Controls/ExtendedAutoCompleteBox.axaml.cs#L157-L199)

The only unsubscribe path is `e.OldValue` inside `OnSuggestionsChanged`, which runs only when the `Suggestions` *property is reassigned*. But `TargetEntityNameSuggestions` and the `Key == "target"` branch both hand out `BspService.Instance.TargetnameIndex.Suggestions` — one collection on a process-lifetime singleton, shared by every I/O row. There's no `OnDetachedFromVisualTree`/`OnUnloaded` anywhere in the file.

When a virtualized row's control is recycled during scroll, the property is never reassigned, so the singleton keeps a strong `CollectionChanged` reference to the dead control and, through its DataContext chain, the whole `EntityPropertyViewModel`/`EntityViewModel` subtree. Scroll a 10,000-entity map a few times and you accumulate thousands of zombie controls; each subsequent throttled rebuild fans `CollectionChanged` out to all of them, each running `UpdateFilteredSuggestions`. The leak costs CPU, not just memory.

Compounding it: `ObservableCollectionExtended`'s `Clear()` and `AddRange()` raise `Action.Reset` with **both** `OldItems` and `NewItems` null, so lines 185-194 unsubscribe nothing and subscribe nothing. After the first targetname rebuild, no item has a handler at all.

### 7. `Dispatcher.UIThread.Post` makes the classname update asynchronous

[`EntityPropertyViewModel.cs:149`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/ViewModels/Shared/Entity/EntityPropertyViewModel.cs#L149)

```csharp
Avalonia.Threading.Dispatcher.UIThread.Post(() => ((EntityViewModel)Parent).Classname = value);
```

- `PullChangesFromModel` does `strVm.Value = strM.Value` for a job-rewritten classname; the `AutoRefresh(Classname)` filter, `MatchClassname`, `PresentableName` and `ClassAndTargetname` all observe the old classname for the rest of that pass.
- `PullChangesFromModel` can be followed by `ResetLumpViewModels()`, which disposes and drops the `EntityViewModel` — the queued lambda then assigns to an orphaned VM and the update is lost.
- In a headless/job/test context nothing pumps the dispatcher, so `Classname` never updates at all.
- The sibling path in the `Key` setter (line 47) still assigns `Classname` synchronously, so the two mutation paths now have different ordering semantics.

This reads as a bandaid for re-entrancy caused by `ValueSuggestions` being a fresh `List` on every get. Memoizing it — exactly the `_inputSuggestions` pattern already used a few lines down — should remove the need for the `Post`.

### 8. `FGDParser.Parse` has no forward-progress guarantee

[`FGDParser.cs:63`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.Generators/FGDGenerator/FGDParser.cs#L63)

```csharp
while (stream.Peek().Type != TokenType.EOF)
{
    if (stream.Match("@")) { ... }   // when false, nothing is consumed
}
```

`ParseEntity` can return parked on a non-`@` token — an entity header with a second description field (`= env_thing : "short" : "long" []`) leaves the stream on `:`, because `ParseEntity` reads only one description before trying `Match("[")`. A leading UTF-8 BOM also lexes as a `Word`, since `char.IsWhiteSpace()` rejects U+FEFF. Symptom: `dotnet build` and the IDE's generator host hang with no error message.

Two more in the same file:

- [Line 216](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.Generators/FGDGenerator/FGDParser.cs#L216): the choices loop does an unguarded `stream.Consume()` for the display token, so `= [ 0 ]` (a choice with no display) swallows the closing bracket and desynchronizes the parser, silently dropping every subsequent entity from `MomentumFGD.Entities`.
- [Line 235](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.Generators/FGDGenerator/FGDParser.cs#L235): `ToDictionary(..., StringComparer.OrdinalIgnoreCase)` throws `ArgumentException` if any choices block has two keys differing only in case.

**None of these fire on the current `momentum.fgd`** — I verified the build is clean. They brick the build on the next FGD edit, which is a nasty failure mode for a checked-in `AdditionalFiles` input.

### 9. `TargetnameIndex` is never disposed, and its subscription defeats lazy lump loading

[`BspService.cs:127`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/Services/BspService.cs#L127)

Two defects at one seam:

1. `BspService.Dispose()` (line 520) disposes `_entityLumpViewModel`, `_pakfileLumpViewModel`, `_bspSubject` and `BspFile` — but not `_targetnameIndex`, which is `IDisposable`. Its subscription stays rooted on a throttled taskpool pipeline that can fire after `_bspSubject` is disposed, calling `Entries.Clear()`/`AddRange` post-teardown.

2. `EntityLumpViewModel` (line 101) is a lazy getter via `LazyLoadLump`, and `EntityLumpViewModelLazy` (line 107) exists *precisely* so callers can peek without triggering construction. `TargetnameIndexViewModel:24` subscribes `BspService.WhenAnyValue(x => x.EntityLumpViewModel)` — the eager getter. `ResetLumpViewModel` nulls the field then raises `PropertyChanged(nameof(EntityLumpViewModel))` (line 478), so once `TargetnameIndex` has been touched once, every reset immediately re-parses and reconstructs the entire entity lump VM, on whatever thread called `ResetLumpViewModels()`. That negates the memory reclaim the reset exists to perform.

### 10. Dropdown button hides forever when suggestions populate asynchronously

[`ExtendedAutoCompleteBox.axaml.cs:183-199`](https://github.com/momentum-mod/lumper/blob/c9f768ea9fc01292a5954cdc54eca6c1c3161e1e/src/Lumper.UI/Controls/ExtendedAutoCompleteBox.axaml.cs#L183-L199)

`SetDropdownButtonState()` sets `_dropdownButton.IsVisible = Suggestions is { Count: > 0 }` and is called from exactly two places: `OnApplyTemplate` and `OnSuggestionsChanged` (property *reassignment*).

`TargetnameIndex.Suggestions` is a single collection instance that starts empty and is filled later via `Clear()`/`AddRange()` on a 200 ms throttle after the BSP loads — the property reference never changes, only its contents. So every Target Entity Name box templated before the index populates evaluates `Count == 0`, hides its dropdown button, and never re-shows it. The user gets a text box with no visible way to open the suggestion list, though keyboard Down-arrow still works.

Add a `SetDropdownButtonState()` call alongside the `UpdateFilteredSuggestions()` in `OnSuggestionsCollectionChanged`.

---

## Non-blocking cleanup

- **`ExtendedAutoCompleteBox` reimplements Avalonia's built-in `AutoCompleteBox`.** Roughly 300 of the 373 lines are re-invented framework behavior: `Text`/`SearchText`, `ItemsSource`, `IsDropDownOpen`, `FilterMode`/`ItemFilter`, `ValueMemberBinding`, popup placement, light dismiss, and Up/Down/Enter/Escape navigation. Genuinely new: `IsBitfieldMode` + the checkbox list, and the dropdown toggle button. Consider hosting or subclassing `AutoCompleteBox` and adding only the bitfield list on top — several of the bugs above (the re-entrancy flags, the focus dispatch hack) are ones the framework control doesn't have.
- **`FGDGenerator.Lit()` should be `SymbolDisplay.FormatLiteral(value, quote: true)`** — already available via the `Microsoft.CodeAnalysis.CSharp` reference. The hand-rolled version escapes only `" \ \n \r \t` and passes every other char through, so a single control character in an FGD description emits invalid C# and breaks the build with an opaque generated-code error.
- **`AddSource("MomentumFGD.g.cs")` is inside a per-value callback** over `IncrementalValuesProvider`, so a second `.fgd` `AdditionalFile` throws a duplicate-hint-name error. Use `.Collect()` and emit once.
- **The incremental pipeline value is the entire ~400 KB file as a `string`.** Roslyn compares pipeline values for equality on every incremental step, so this is a full-buffer compare plus a fresh LOH allocation per keystroke in the IDE. Project to a checksum first and read the text in the downstream stage.
- **`ValueSuggestions` allocates a fresh `List` and fresh items on every get** — for `Key == "classname"` that's one object per FGD entity, several hundred, on every read. `InputSuggestions` right below it already memoizes; do the same, and hoist the classname list to a `static readonly` since it's identical for every row.
- **`FetchInputSuggestions` does an O(n) `Entries.FirstOrDefault(...)` scan** over a 1k-5k element list. The `GroupBy` in `TargetnameIndexViewModel` already produces exactly the `targetname → classname` shape a `Dictionary` would want.
- **`UpdateFilteredSuggestions` does `Clear()` then `Add` in a loop** into an `ObservableCollection` that is the live `ItemsSource` of two `ListBox`es — one `CollectionChanged` and one layout invalidation per item, on every keystroke.
- **`OnClassnameChanged()` is dead** — declared, never called, never overridden.
- **`BspService.Instance.TargetnameIndex?.Suggestions ?? []`** guards against a null that can't occur; `TargetnameIndex` is non-nullable and lazily constructed. Two other call sites in the same file deref it unguarded.
- **`ExtendedAutoCompleteItem` hand-rolls `INotifyPropertyChanged`** — it's the only INPC implementation in the repo that isn't a `ReactiveObject`. `[Reactive]` on the three properties would delete all of it.
- **`Parent` vs `ParentEntity`.** The diff adds a strongly-typed `ParentEntity`, but the file now has three spellings of the same reference: `ParentEntity`, `((EntityViewModel)Parent)`, and `Parent is EntityViewModel entityVm`. Worth mechanically collapsing to the new one.
- **Magic strings over the FGD type system.** `"classname"`, `"targetname"`, `"target"`, `"flags"`, `"boolean"` appear 15 times across two files. `Key == "target"` in particular special-cases one key name when the FGD already declares a `target_destination` `ValueType` — dispatching on `ValueType` would cover `filtername`, `parentname`, and the rest for free.
- **`Lumper.sln`:** 50 added lines declare `Debug|x86`/`x64` and `Release|x86`/`x64` configurations that all map straight back to `Any CPU`, for a project set where nothing is platform-specific.
- **`Lumper.Lib.csproj`:** `EmitCompilerGeneratedFiles=false` means nothing is ever written to disk, so `CompilerGeneratedFilesOutputPath` is inert and the `<Compile Remove>` guarded on it is a no-op. Pick one intent.

Natanxp2:
I went through all the review points from opus. Addressed all of them except for two:

### ExtendedAutoCompleteBox reimplements Avalonia's built-in AutoCompleteBox
I specifically avoided extending original autocompletebox because it sounds like an absolute nightmare.

### ExtendedAutoCompleteItem hand-rolls INotifyPropertyChanged
Using [Reactive] here introduces insane performance issues when loading/scrolling. I have no idea why

tsa96 (approved):

